Issue #46, opened 2026-08-05
Phase C: `spec-correctness` skill — generate property-based tests from acceptance criteria


kreneskyp:
## Background

**quoin** is the specification workflow toolkit for this ecosystem. Its skills
live in `skills/*` and are invoked as slash commands — `spec-matrix` builds a
repo's Test Matrix (`spec/tests.md`), `gap-analysis` verifies that the matrix is
backed by real tests and that code traces to spec, `specify` authors requirement
artifacts, and so on.

An **acceptance criterion** in this ecosystem is a row in a requirement
document's `## Acceptance Criteria` table, with columns
`ID | Criteria | Verification`. The link from a criterion to the test that
verifies it is currently:

- a free-text annotation in the `Verification` cell — `Test (TC-035)`, and
- a `Traces To` column in the Test Matrix,

reconciled against grep-able tracking tags in test code by an agent step inside
`gap-analysis`. The engine side of this is being replaced by declarative
machinery (quire-rs FR-049 checks that `TC-035` exists; FR-050 computes the
rollup; FR-051 extracts source symbols and framework-native trace markers).

**What does not exist anywhere is generation.** No skill turns an acceptance
criterion into a test. Every `Type = Property` row in the ecosystem is
hand-authored.

## What this issue asks for

A new quoin skill, `spec-correctness`, peer to `spec-matrix` and `gap-analysis`,
implementing a **property-based test generation loop** from acceptance criteria:

- **Python** → Hypothesis
- **Rust** → proptest
- **TypeScript** → fast-check

Property-based testing means generating many inputs and asserting a property
holds over all of them, with automatic shrinking to a minimal counterexample on
failure. It is a much stronger oracle than an example test for criteria that are
universally quantified — and roughly a quarter of ecosystem acceptance criteria
already are (50.8% in quire-rs itself).

## Required shape of the loop

This is not a one-shot "LLM writes a property from a sentence" task. Published
work on LLM-generated properties is consistent that one-shot generation is
unreliable in both validity and soundness, and that the agentic loop is what
makes it trustworthy. The loop is therefore:

1. **Precondition** — the spec validates clean, and the criterion has been
   classified as property-shaped by the engine (see Dependencies).
2. **Generate** — write the property test for the criterion, emitting the AC
   tracking tag into the test so the existing `gap-analysis` reconciliation picks
   it up unchanged. Do not invent a new tagging convention.
3. **Execute** — actually run it.
4. **Reflect and triage every failure three ways:**
[...]

PR #58, opened 2026-08-08
feat(skills): spec-correctness — property tests from classified criteria

kreneskyp:
Phase C of the spec-PBT program. Closes the loop opened by agent-ix/quoin#46.

quire-rs FR-052 classifies every binding acceptance criterion by property shape and stops there — it names no test framework (CON-2) and its deterministic recall is a measured ceiling of 31.3% at 93.3% precision (quire-rs#45). This adds the consumer that closes the three gaps that leaves: the harness mapping, the clause grounding, and the recall residue.

## Spec first

- `FR-028` + `US-011`, traced to `ix://agent-ix/quire-rs/FR-052`.
- `EV-050`..`EV-053` define the eval coverage. FR-028 and US-011 are marked **spec-ahead-of-coverage** in `spec/matrix.md` until those scenarios are implemented in `evals/scenarios/` — the row does not claim coverage it does not have.

## The skill

`skills/spec-correctness/` — SKILL.md, eight step references, a strategy table, and one real template per harness (proptest / fast-check / hypothesis).

- **Routing** keys on `extraction`; **strategy** keys on `property`. `shape` is the FR-047 grammar axis and selects nothing.
- Tests carry the criterion's `row_id` in **two independent carriers**, so `gap-analysis` reconciles a generated test by the same grep as a hand-written one.
- Queued tests are **inert by construction** — an unreviewed test cannot turn a matrix row green, and acceptance leaves the tracking tag byte-identical.
- CON-1 throughout: a low extractable ratio is data, never a verdict, and nothing suggests rewording a criterion.

## Dogfooding corrected four things

Running it on quoin and quire-rs before shipping it found real bugs in the design:

1. A file importing an **absent generator library breaks test collection** even when every test in it is skipped. quoin has no `fast-check`, so the run writes proposals as code blocks and installs nothing.
2. **Cargo auto-discovers integration tests only at `tests/*.rs`** — the original nested `tests/props/fr_NNN.rs` layout would never have run.
3. Rust imports use the **`[lib] name`**, which routinely differs from the package name (`quire-rs` → `quire_rs`).
4. **`property: universal` over-counts.** English writes a universal and a single witness with the same determiner, so grounding decides which a criterion is — 31 of quoin's 52 are witnesses. quire-rs added two more: a metamorphic label can come from a criterion *naming* a shape, and a criterion can be written as a description of its own test.

None of these is a classifier defect and none is a reason to reword a criterion.

## Evidence

- `tests/props/QUEUE.md` — the quoin run: 130 criteria, 17 grounded property proposals, 31 witnesses, 12 refusals, 70 left to the second pass. Counts reconcile to the record total.
- agent-ix/quire-rs PR (branch `feat/property-tests-fr-040`) — four **generated, passing** proptest properties over `normalize_allowed_links`, plus the CR-034 note recording FR-052's first consumer.

## Verification

- `make lint` clean; `make test` 179 passed.
- `quire validate --scope . 'spec/**/*.md'` — no new findings (all remaining EARS warnings pre-date this branch).
- quire-rs: `cargo test --test props_fr_040` 4/4 pass, `cargo fmt --check` and `cargo clippy --tests` clean.

## Follow-ups, not blockers

- Implement `EV-050`..`EV-053` in `evals/scenarios/`, which flips FR-028 and US-011 off spec-ahead-of-coverage.
- Add a `spec-correctness` value to `SpecReview.analysis` in `spec-artifacts-process` if the queue should ever become a validated review artifact. Until then it lives in the test tree, outside every spec validation glob.
- Add `fast-check` to quoin's devDependencies to emit the 17 queued proposals as real tests.


kreneskyp:
## Updated — fast-check added, 17 tests emitted, EV-050..EV-053 implemented

Two commits since the PR opened.

**`test(props)` — the dogfood is now real coverage, not proposals.** Added `fast-check` and re-ran the skill: **17 criteria emitted unattended** across FR-002, FR-005, FR-006, FR-008, FR-009, FR-010, FR-012, FR-013, FR-015, FR-018 and FR-025, each carrying its `row_id` in two carriers. **204 tests pass and the 100% coverage gate still holds.**

The emit caught two bugs, both in the generated tests rather than in `src/` — which is exactly what step 4 says to expect from a failing unattended test:

- four catalog properties built fixtures with `artifacts:`/`objects:` where `loadCatalog` reads **`artifact_types:`/`object_types:`**, so the oracle was right and the domain was empty;
- `FR-005-AC-1`'s second assertion used a `\b`-anchored regex on a generated command name. fast-check shrank it to `"a-"`, where `\b` does not match after a hyphen.

Both harness-default deviations (`numRuns` 30 for the oclif-dispatching FR-005 properties, 40 for the FR-015 properties that create a directory per case) are recorded in `QUEUE.md` with their reasons.

**`test(evals)` — FR-028 and US-011 are off spec-ahead-of-coverage.** EV-050..EV-053 are implemented in `evals/scenarios/index.mjs`. The fixture criteria were **checked against the real classifier rather than assumed**:

| scenario | fixture classification |
| --- | --- |
| EV-050..EV-052 | `universal` + `ordering` + `error-case` all extractable, plus one `example` witness that must never produce an unattended test |
| EV-053 | two criteria the classifier calls **extractable** but grounding must **refuse** — one adjectival oracle, one absent symbol — and a manifest with no generator library |

EV-053 asserts the negative too: no `verdict`, no `reword` anywhere in the report.

Ready for review. Release remains user-gated.

Issue #65, opened 2026-08-13
Realign EV- ids to TC-, and 55 status lies now that coverage can see the matrix


kreneskyp:
`quire coverage` reported `0/0 rows backed (100%)` here until agent-ix/spec-artifacts-process#22 shipped a trace-tag declaration and #61 fixed the tags it could not bind. Current:

```
[RAN] quoin   94/273 rows backed   status_lies=55   untracked_symbols=1

  acceptance-criterion       94/123   (was 55/123 before #61)
  test-case (spec/matrix.md)  0/109
  test-case (spec/evals.md)   0/40
```

Pre-existing. Filed rather than folded into the #61/#62/#63 PR.

## 1. Collapse the `EV-` / `TC-EV-` shadow

**[RAN]** `spec/evals.md` carries **40 bare `EV-nnn` ids and 40 `TC-EV-nnn` ids** — the same evals, numbered twice. The matrix rows use `TC-EV-nnn` (already CR-019-compliant); the *scenario* id used by `evals/scenarios/index.mjs` and by FR `Verification` cells (`Eval (EV-050)`) is the bare form.

This is settled, not open. CR-019 renamed `BENCH-`, `AUDIT-`, `SB-` and `IS-` to `TC-` because a prefix naming no declared archetype "encodes in the id what the row already states one column over". `EV-` is that case and was missed.

The enabling half has landed: **`Eval` is now a `test_type` value** (spec-artifacts-process CR-020). Before that, the rows had nowhere to move the information to and sat at `Type: E2E`, which is not what an agent-behaviour eval is.

Work:

- `EV-nnn` → `TC-EV-nnn` in `evals/scenarios/index.mjs` (64 refs), `spec/matrix.md` (14), and the FR `Verification` cells that read `Eval (EV-050)`.
- Flip those rows' `Type` from `E2E` to `Eval`.
- Drop the shadow column from `spec/evals.md` so each eval has exactly one id.

Concrete payoff: a `Verification` cell reading `Eval (EV-050)` extracts no test id under the FR-004 reference patterns, which admit `TC`/`IT` only — so those rows fall back to their own criterion id. After the rename they bind directly.

## 2. The 55 status lies

**Acceptance criteria are in good order** — 94 of 123, and every `row_id` the `spec-correctness` run emitted is backed.

[...]

Issue #68, opened 2026-08-14
release-drift only diffs src/, so a stale module pin ships silently


kreneskyp:
`release-drift.yml` guards against "code merged to main that never gets released". It does that by diffing `src/`:

```
if git diff --quiet "$LAST_TAG"..HEAD -- src; then
```

`default-modules.yaml` is at the **package root**, not under `src/`. So the guard is structurally blind to the file that decides which module version every fresh install receives.

## It already cost a release

**[RAN]** v0.12.2 shipped with `spec-artifacts-process` pinned at **v0.10.0** while its own skills required v0.11.0. That pin has no `trace_tags`, no `document_references`, and no `spec-correctness` value in the `SpecReview.analysis` enum, so for a fresh install v0.12.2 failed in both directions at once:

- `quire coverage` backed zero rows and reported `0/0 (100%)` — the exact failure 0.12.2 exists to fix;
- the `spec-correctness` skill could not write its own artifact, because `analysis: spec-correctness` fails validation against v0.10.0.

Fixed in #67, shipped as v0.12.3.

## Why no local check caught it either

The verifying machine had the module updated by hand to v0.11.0 before the release was checked. Every test — including an end-to-end `quire coverage --scope .` with no `--module` override — passed against a state no new user has. **A developer machine is not a representative environment for a lazily-installed default module set**, and nothing forced the check into a clean one.

## Work

Two independent gaps, either of which would have caught it:

1. **Widen the drift guard** beyond `src/` to the files that affect published behaviour — at minimum `default-modules.yaml`, `package.json`, `bin/`, `skills/`. Simplest correct form is to diff everything the package actually ships (the `files` list) rather than enumerate paths that will drift again.

2. **Assert the pins are current.** A check that reads `default-modules.yaml` and compares each `ref` against that repo's latest tag. It should *report*, not auto-fail: pinning behind latest is sometimes deliberate, and the value is knowing rather than being forced. All eight pins were audited by hand for #67 — seven were current, which is exactly the kind of result a script should produce, not a person.

3. Consider a smoke test that runs `module ensure-defaults` against a **clean `IX_CONFIG_ROOT`** and asserts the resolved module set matches the manifest. `install-smoke.yml` already exists and is the natural home; the fresh-root form is what makes it representative.

PR #73, opened 2026-08-14
Split comma-listed Trace comments so every id binds

kreneskyp:
Closes the taggable half of #65. **Stacked on #72.**

## All 15 matrix lies had one cause, and it was not the matrix

The declared `typescript-trace-line` legacy form captures **one** id per comment:

```
Trace:\s*([A-Za-z]+-\d+(?:-[A-Za-z]+-\d+)?)
```

So a test written

```ts
// Trace: FR-001-AC-1, FR-001-AC-2, FR-001-AC-4
```

bound `FR-001-AC-1` and **silently dropped the other two**. The tags are right there in the file — greppable, readable, and invisible to the rollup. That is the same failure shape as the Rust raw-string defect in agent-ix/quire-rs#65: a tag that looks present and binds to nothing.

**24 ids across 6 test files were binding to nothing.** Splitting them onto their own lines changes no claim; it makes the existing one visible.

## Measured

```
backed rows    94/273 → 109/273
status lies    55 → 40
matrix lies    15 → 0
```

## What the remaining 40 are

Every one is a `spec/evals.md` row — agent-behaviour evals with no source symbol to tag. They are exactly what quire-rs **CR-041**'s `no_source_symbol` vocabulary exists for, tracked in agent-ix/spec-artifacts-process#26. No further comment fixes them, and none of them is an overclaim.

Once that lands, quoin's status lies go to **0**.

## Verification

- `make lint` clean, `make test` 225/225
- Coverage measured before and after with the installed `quire` CLI


kreneskyp:
## This merges as-is; the grammar fix is filed separately

Filed **agent-ix/quire-rs#68** to support the comma-separated form properly, rather than leaving the corpus to keep writing something that reads as several ids and means one.

Two things the measurement for that ticket turned up, which change how this PR should be read:

**1. It is an engine asymmetry, not a pattern typo.** `src/symbols/trace.rs` splits capture group 1 for canonical **markers** (`marker_ids`) and returns it whole for **legacy** forms (`legacy_id`). So `#[trace("TC-001", "FR-007-AC-1")]` binds both ids and `// Trace: A, B` binds one. Widening the module regex alone cannot fix it — `legacy_id` would return `"A, B"` as a single id and resolve nothing.

**2. quoin's 24 were a tenth of it.** Across `~/dev`, worktrees excluded:

```
legacy comment lines carrying a list:  98
ids that bind to nothing today:       205
repos affected:                        17
```

and it is every language, not just TypeScript — `//!` in Rust, `"""` in Python, `*` in JSDoc:

```
config-service/tests/test_service.py     """FR-007-AC-1, FR-005-AC-1
filament-ide-rs/.../reconciliation.rs    //! TC-058, TC-198
ticket-runner/tests/control-guarantees.test.ts  * Trace: FR-013, FR-014
```

`ticket-runner` alone drops 84 ids; `filament-ide-rs` drops 62.

**This PR still stands on its own.** Splitting quoin's comments is the correct repo-local fix, it takes matrix lies 15 → 0 today, and it stays correct after #68 lands — one id per line is unambiguous either way. #68 is what stops the next repo hitting this.

PR #117, opened 2026-08-18
feat: pin spec-objects-safety as the ninth default module

kreneskyp:
Last piece of agent-ix/spec-objects-security#7. **Draft — blocked on a visibility decision, see below.**

Adds `spec-objects-safety` (hazard, failure_mode) to the default module set, pinned at `v0.2.0` like every other entry.

## Blocked: the module repo is private

`git-subdir` pins clone from the source repo. **`agent-ix/spec-objects-safety` is PRIVATE; all five sibling `spec-objects-*` repos are PUBLIC.** Until it matches, this pin fails to install for anyone without repo access.

I created it private because that is the standing default for new repos. The family convention is public — they publish to public npm as `@agent-ix/*` and their pins are cloned by every consumer. **Someone needs to decide which applies here** before this merges.

## The guard did its job

`tests/index.test.ts` asserts the module set has an **exact** count, not a lower bound. Adding an entry failed it, and updating the count cost a deliberate line — which is the right shape, because this set installs into every consumer's `~/.ix/filament/modules`. A silent addition is a silent change to everyone's catalog. Updated 8 → 9, with `spec-objects-safety` added to the name assertion too.

## Gates

`make build` ✅ · `make test` ✅ 329 passed · `make lint` ✅

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,3 +13,23 @@ make add-dev-packages p=<name>  # add dev dependency
 make use-local p=<name>         # switch dep to local package
 make use-upstream p=<name>      # switch dep back to upstream
 ```
+
+## Adding or improving a spec check
+
+See **CLAUDE.md § Adding or improving a spec check**. The short form:
+
+A new check pointed at the `~/dev` corpus will fire in the hundreds or thousands.
+**That is expected, not evidence the check is wrong.** A high count means one of
+two things, settled by reading flagged documents, not by preference:
+
+- **Bad rule** — the check misreads correct data.
+- **Bad corpus** — the check reads correctly and the specs are wrong.
+
+Do not default to either. Agents wrote most of these specs and agents do not
+write good specs — that is why quoin and quire exist.
+
+**Never widen a rule because it lowers the count.** A rule states what a good spec
+looks like; it does not fit the specs that exist. Where two forms mean the same
+thing, prefer unifying the corpus on one and flagging the rest over accepting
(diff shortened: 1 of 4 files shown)


kreneskyp:
## Visibility blocker resolved

The draft was held on one question: `git-subdir` pins clone from the source repo, and `agent-ix/spec-objects-safety` was **private** while all five sibling `spec-objects-*` repos are public. A private module fails to install for anyone without repo access.

Resolved: the repo is now **public under AGPL-3.0-or-later**, matching the family.

Two follow-on changes:

1. `pyproject.toml` had **no license field at all** — the cookiecutter never set one. `LICENSE` (AGPL v3) and `package.json` already carried it, but the wheel's declared license comes from `pyproject.toml`, so a build from `v0.2.0` would publish with no license metadata. Fixed and released as **`v0.3.0`**; this pin moves to it.
2. GitHub now detects the license as `agpl-3.0` on the repo.

**Verified, not assumed** — cloned the pinned tag with credentials stripped, so it exercises the anonymous path a fresh install takes:

```
GIT_TERMINAL_PROMPT=0 GIT_CONFIG_GLOBAL=/dev/null \
  git clone --depth 1 --branch v0.3.0 https://github.com/agent-ix/spec-objects-safety.git
```

Succeeds; `spec_objects_safety/manifest.yaml` and `skeletons/` are present.

Gates re-run after the bump: `make build` ✅, `make test` ✅ 329 passed / 30 files, `make lint` ✅.

PR #186, opened 2026-08-21
fix(review): include EARS in the enforced all set

kreneskyp:
## Summary
- include ears-conformance in the workflow-backed all review set
- align workflow instructions, intake, reference template, and acceptance invariant with the direct skill
- add a vocabulary-drift regression test across all declarations and the installed SpecReview schema
- prove acceptance blocks until the EARS review document is recorded

## Verification
- make test: 47 files, 524 tests passed
- make lint: passed
- Quire specification validation: passed with existing advisory warnings

## Note
The build completes with existing declaration diagnostics in unrelated evidence-store and SBOM sources.

### DIFF
--- a/skills/spec-review/workflow-assets/skills/review/SKILL.md
+++ b/skills/spec-review/workflow-assets/skills/review/SKILL.md
@@ -24,9 +24,9 @@ document per analysis** under `spec/reviews/`, rather than a single freeform rep
   record it in the `request` interview as `review_set` plus `selected_analyses`:
   - `base` — the skill checklist only (ID formats, US/FR/TC quality, the six coverage
     rules). No analysis skills. `selected_analyses` is empty.
-  - `all` — `base` plus all six analyses: `failure-domain`, `integrity`, `dependency`,
-    `evidence`, `risk-complexity`, `scope-boundary`.
-  - `subset` — `base` plus the analyses the user picks from those six.
+  - `all` — `base` plus all seven analyses: `failure-domain`, `integrity`, `dependency`,
+    `evidence`, `risk-complexity`, `scope-boundary`, `ears-conformance`.
+  - `subset` — `base` plus the analyses the user picks from those seven.
 - **analyses_run — run the selected set, one SpecReview doc each.** Run each selected
   analysis (the matching `spec-*-analysis` skill). Prefer running them **in parallel** —
   each writes its own file, so there is no contention.
@@ -55,7 +55,7 @@ hand-copy). The structure below is a reference for what that skeleton contains �
 id: SR-001
 title: "<analysis> review of <scope>"
 type: SpecReview
-analysis: <failure-domain|integrity|dependency|evidence|risk-complexity|scope-boundary|base>
+analysis: <failure-domain|integrity|dependency|evidence|risk-complexity|scope-boundary|ears-conformance|base>
 scope: <spec paths / ids>
 review_set: <base|all|subset>
 ---
--- a/skills/spec-review/workflow-assets/skills/review/scripts/invariants.js
+++ b/skills/spec-review/workflow-assets/skills/review/scripts/invariants.js
@@ -1,6 +1,6 @@
 import { specInvariants } from "../../../dist/index.js";
 
-// Canonical "all" set — the six analyses named in the spec-review skill.
+// Canonical "all" set — the seven analyses named in the spec-review skill.
 const ALL_ANALYSES = [
   "failure-domain",
   "integrity",
@@ -8,6 +8,7 @@ const ALL_ANALYSES = [
   "evidence",
   "risk-complexity",
   "scope-boundary",
+  "ears-conformance",
 ];
 
 const norm = (value) =>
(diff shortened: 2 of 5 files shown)


kreneskyp (commented):
## Code review + gap analysis — head `07aa6995` (draft)

Reviewed the diff, the workflow assets, and the new test against `origin/main`. Gates were **not** executed here — findings are from reading, plus the `git show origin/main` checks quoted below.

Clean on the identity check that blocks the other seven drafts in this repo: this PR mints no new requirement id, it edits `US-007`. The drift it fixes is real — `origin/main`'s `skills/spec-review/SKILL.md` has carried seven analysis rows including `ears-conformance` since before this branch, while the workflow assets, the intake prompt and the invariant all still said six. A vocabulary declared in one place and enforced from another is exactly the shape worth a test.

**Verdict: CONDITIONAL** — one environment dependency that can turn the new gate into a red build for a reason unrelated to drift, and one traceability gap.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | medium | The schema half of the drift test reads an ambient `~/.ix` install, so it fails or passes for reasons outside this repo | `tests/spec-review-vocab-drift.test.ts` |
| FND-002 | medium | A 156-line test with two real gates carries no TC id and gets no matrix row | `tests/spec-review-vocab-drift.test.ts`, `spec/matrix.md` |
| FND-003 | low | `US-007-EX-4` is added under a heading that declares its own examples non-verifying | `spec/usecase/US-007-review-into-specreview-docs.md` |

### FND-001 — the one non-hermetic assertion in an otherwise hermetic test

Five of the six vocabulary sources are read from files inside this repo. The sixth is not:

```ts
const schema = JSON.parse(
  readFileSync(
    join(filamentModulesDir(), "spec-artifacts-process/schemas/spec-review-frontmatter.schema.json"),
    "utf8",
  ),
) as { properties: { analysis: { enum: string[] } } };
for (const analysis of direct) {
  expect(schema.properties.analysis.enum).toContain(analysis);
}
```

`filamentModulesDir()` resolves to `join(ixHome(), "filament", "modules")` — a machine-local install directory. Quoin ships no copy of that schema; `src/quire/schemas/` holds only `coverage-v1` and `properties-v1`.

Three ways this goes wrong, none of which is a vocabulary drift:

1. **The module is not installed.** `readFileSync` throws ENOENT and the whole `it` block errors. The reported failure names a missing path, not a drift, and the other five comparisons in the same `it` never run — so a genuine drift introduced on the same commit is masked by the environment error.
2. **An older `spec-artifacts-process` is installed.** Its `analysis` enum lacks `ears-conformance`, the test goes red, and the diagnosis is "your machine is stale", not "the repo drifted". Given this repo already has a documented history of tests gated on an unpublished dependency version, that is not hypothetical.
3. **A newer one is installed.** Its enum contains everything, so the assertion passes regardless of what the five in-repo sources say. In that direction the check contributes nothing.

The other five comparisons are the valuable part and they are fully hermetic. Suggest either pinning a vendored copy of the enum the way `src/quire/schemas/` pins the Quire contract, or splitting the schema check into its own `it` that skips explicitly — with a stated reason — when the module is absent. A skipped check that says why is honest; one that errors on ENOENT reads as a failure of this PR.

### FND-002 — an untraced gate

`tests/spec-review-vocab-drift.test.ts` is 156 lines and contains two genuinely load-bearing assertions: the six-way vocabulary comparison, and a dynamic import of `invariants.js` that proves an `all`-set run stays blocked with `missing: ["ears-conformance"]` until the document is recorded. Neither carries a `// TC-NNN` marker, the describe title carries no id, and this PR adds no row to `spec/matrix.md`. `origin/main`'s matrix has no `US-007` rows either, so nothing in the matrix points at this file from any direction.

Failure scenario: someone deletes the second `it` while trimming a slow suite. No matrix row goes stale, no coverage figure moves, and the gate that enforces `ears-conformance` at the workflow boundary is gone silently. This repo's convention elsewhere — `tests/assurance.test.ts` on main — is a header range plus per-test `// TC-261` markers; worth following here.

### FND-003 — the example is filed where examples do not verify

`US-007-EX-4` is added under a heading whose own text reads "These examples clarify expectations; they are illustrative, not verification criteria." So the one place the new behaviour is written down as a requirement explicitly disclaims being one, and FND-002 means no matrix row picks it up either. Either promote it to an acceptance criterion on the owning requirement, or add the TC row so the test is the binding record.

### Verified good

- The drift this catches was real and had already shipped: `git show origin/main:skills/spec-review/SKILL.md` lists all seven analyses while `invariants.js`'s `ALL_ANALYSES` listed six, so an `all`-set run would have accepted itself without an EARS review document. That is a gate passing because it could not see the thing it was meant to check — the same class as the findings on #275 and #381 — and this is the fix for it.
- Bumping `def.yaml` from `0.2.0` to `0.2.1` alongside the prompt change means a persisted in-flight workflow instance can tell which vocabulary it was started under.
- Comparing all five in-repo sources against each other, rather than each against a hard-coded list, means the test keeps working when an eighth analysis is added — only `expect(direct.size).toBe(7)` needs touching, and that one line failing is a clear, correct signal.
- The invariant test drives the real `selected_analyses_covered` function through a dynamic import rather than reimplementing its logic, and asserts on the structured `details.missing` payload rather than on a message string.

kreneskyp:
Superseded by `main`. The EARS check is already in the enforced set: `skills/spec-review/workflow-assets/skills/review/scripts/invariants.js:11` carries `"ears-conformance"`.

Closing as part of the pre-Wave-3 stabilization sweep. This PR was opened 2026-08-21; `main` has since advanced from FR-042 to FR-092, and the spec ids this branch writes (FR-043..FR-047) now belong to entirely different requirements. Nothing here is lost — the behaviour landed by another route.

The workflow-integration work that was stacked on this branch (#194) has been retargeted onto `main` and is being landed separately.

PR #187, opened 2026-08-21
feat(evidence): add generic measurement record

kreneskyp:
## Summary
- specify MeasurementRecord as a third, policy-free evidence shape distinct from run and finding records
- publish and enforce a closed v1 JSON Schema with plan/definition, subject/scope/revision, value/unit, distribution, and provenance identity
- persist validated records canonically under spec/evidence/measurements and ship the schema in built packages
- cover CLI latency distributions and per-function complexity observations plus invalid identity, non-finite values, stable paths, round trips, and policy exclusion

## Validation
- make test: 47 files, 534 tests passed
- focused measurement coverage: 100% statements/branches/functions/lines
- Quire validation passed (existing duplicate-module advisories only)
- build completed and copied measurement-record-v1.schema.json; existing unrelated declaration diagnostics in affirm.ts, record.ts, and sbom.ts remain

## Scope
Thresholds, gates, aggregation policy, dashboards, and a universal quality score are intentionally deferred to the comparison layer.


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-043, which is already taken:**

- `spec/functional/FR-043-generic-measurement-record.md`

quoin#194 also mints an FR-043 (`FR-043-assurance-aware-spec-workflows.md`), so this id is claimed three ways: `main`, this PR, and a sibling draft.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `6d52381d` (draft)

Reviewed `src/evidence/measurement.ts`, the schema, the store additions, FR-043 and the tests. Gates were **not** executed here — findings are from reading. The FR-043 id collision is covered in my earlier comment and is not repeated below.

The record shape is well judged: policy-free, measure-neutral, closed to undeclared fields, and the doc comment on `MeasurementRecord` does the useful work of saying what it deliberately is *not*. Every TC id in the matrix row appears in a describe title, which is better than the sibling PRs in this stack. The findings below are about identity and about how strong the digests actually are.

**Verdict: CONDITIONAL** — one gap where the requirement itself is what needs changing, not the code.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | Record identity omits tool, environment and sampling, so a re-measure silently overwrites | `src/evidence/store.ts` (`measurementPath`), FR-043-AC-4 |
| FND-002 | medium | The `digest` pattern accepts any algorithm and any length, down to one hex character | `src/evidence/schemas/measurement-record-v1.schema.json` (`$defs/digest`) |
| FND-003 | medium | Overwrite/replacement has no stated requirement and no test | FR-043 acceptance criteria |
| FND-004 | low | `readMeasurement` throws on a corrupt file, so one bad record kills a bulk read | `src/evidence/store.ts` (`readMeasurement`) |
| FND-005 | low | `sourceRevision` is `minLength: 1` while `collectedAt` is tightly patterned | schema |

### FND-001 — the store cannot hold two environments for one observation

```ts
const identity = canonicalJson({
  definitionVersion: validated.plan.definitionVersion,
  scope: validated.scope,
  sourceRevision: validated.sourceRevision,
  subject: validated.subject,
});
```

and FR-043-AC-4 states the same set as the requirement: "a changed subject, scope, revision, or definition version changes that path."

So `tool`, `environment` and `sampling` are persisted but are not part of the record's identity. Same plan, same definition version, same subject, same scope, same revision, measured twice with anything else different, resolves to one path — and `writeMeasurement` calls `writeCanonical` unconditionally, so the second write replaces the first with no record that it happened.

Failure scenario, using this PR's own fixture: `measurement-cli-latency.json` collected at commit `abc123` on CI (`environment.id: "ci"`, `sampling.sampleCount: 1000`), then the same measure run on a developer laptop at the same commit (`environment.id: "laptop"`, `sampleCount: 10`). The laptop number lands on top of the CI number. The comparison layer built on this in #188 then either reports a regression that is really two machines, or reports the machine you did not ask about, and there is nothing in the store to say a value was replaced.

The same happens across `tool.version` — dependency-cruiser 15 then 16 at one revision — which is exactly the comparison a code-health measure exists to make.

That the record carries `environment.id` and `sampling.id` as first-class fields is the tell: they are there because measurements vary by them. An identity that excludes them cannot represent that variation.

This is a requirement finding as much as a code one. AC-4 verifies the current behaviour correctly; it is the enumerated set in AC-4 that needs widening to include tool, environment and sampling identity, and then `measurementPath` follows. If some of those are deliberately meant to be *replaced* rather than retained — a re-run with a fixed config superseding a bad one — that is a legitimate design, but it needs to be stated and it needs FND-003.

### FND-002 — an integrity anchor with no strength

```json
"digest": {
  "type": "string",
  "pattern": "^[A-Za-z0-9][A-Za-z0-9._+-]*:[A-Fa-f0-9]+$"
}
```

This guards both `tool.configurationDigest` and `rawEvidence.digest` — the two fields that make the observation traceable to the artifact and configuration it came from. The pattern constrains the shape and nothing else:

- **Any length.** `sha256:a` validates. A producer whose digest computation fails and emits a truncated or placeholder value writes a record that passes validation, and every consumer downstream treats it as a distinct, verified reference.
- **Any algorithm.** `md5:...` and `crc32:...` validate equally. For a field whose job is integrity, the algorithm is not an implementation detail.
- **Mixed case.** `A-Fa-f` means `sha256:ABC` and `sha256:abc` are two different strings for the same digest, so two producers that disagree on casing produce records that will never compare equal.

Compare quoin#275, which pins its attestation digest to `^sha256:[0-9a-f]{64}$` — the same repo, the same kind of field, an order of magnitude stronger. Either an allow-list of `(algorithm, hex length)` pairs, or at minimum a per-algorithm `minLength`, plus lowercase-only.

### FND-003 — nothing states what a second write means

AC-1 through AC-6 cover shape, rejection, non-finite refusal, path determinism, provenance round-trip and schema closure. None of them says what happens when a record is written to a path that already holds one.

That is the unstated requirement behind FND-001, and it is the one a reader of this store most needs. Three defensible answers — last-write-wins, refuse-if-different, append-a-new-observation — and the code silently picks the first. Whichever you choose, it wants an AC and a test; right now a future change from overwrite to refusal would break no test and violate no stated criterion.

### FND-004 — one corrupt file stops a bulk read

```ts
export function readMeasurement(path: string): MeasurementRecord | null {
  const record = readJson<unknown>(path);
  return record === null ? null : validateMeasurementRecord(record);
}
```

Absence returns `null`; corruption throws. That is a reasonable split for a single-record read, but the comparison layer will walk `measurements/<plan>/` in bulk, and there one hand-edited or half-written file takes down the whole scan with a `MeasurementValidationError` naming a path.

`readBundleFrontmatter` in the sibling #189 already models the alternative: collect an `unreadable: Array<{path, reason}>` and continue, so the report says which records it could not read *and* still answers for the ones it could. Worth adopting before #188 depends on this.

### FND-005 — one identity field is patterned, another is not

`collectedAt` is pinned to UTC-with-`Z`, which is right and matches the doc comment. `sourceRevision` is `minLength: 1`, so a branch name, a tag, a short SHA and a full SHA are four different identities for one code state — and `sourceRevision` is *in* the path digest, so each of them gets its own file. Since AC-4 makes revision an identity component, it deserves the same treatment `collectedAt` got.

### Verified good

- `strictNumbers: true` with a comment saying exactly why ("JSON.stringify silently converts NaN and infinities to null, so the record must reject them before persistence") is the kind of guard that is invisible until it saves you, and AC-3 pins it at every numeric position rather than just the primary value.
- Validating inside `measurementPath` means the path cannot be computed for a record that would not survive a write — the ordering is right, and it is why `subject.id` and `sourceRevision` can never become path segments.
- `additionalProperties: false` at every level of the schema, including nested `quantiles` items. This is the posture quire-rs#381's clause-set input types are missing.
- `distribution.anyOf` requiring at least one of minimum/maximum/mean/standardDeviation/quantiles alongside `count` stops a distribution object that carries a count and says nothing.
- All six TC ids appear in their describe titles, so the matrix row binds to real tests. Worth keeping when #190 and #186 are fixed.

### One smaller note

`distribution` has no cross-field constraints: `minimum` may exceed `maximum`, and `quantiles` has no `uniqueItems`, so two entries with the same `probability` and different `value` both validate. A duplicate p95 with conflicting values survives to the comparison layer. `uniqueItems` on the array is one line; the min/max ordering needs a code check since JSON Schema cannot express it.

kreneskyp:
Superseded by `main`. The generic measurement record landed by another route: `src/evidence/record.ts` exports `RecordRequest`, `RecordOutcome`, `recordRun()` and `obligationsFrom()`, backed by a 10-adapter registry under `src/evidence/adapters/` (junit, sarif, sbom, cargo-mutants, contract-conformance, agent-eval, audit-script, differential-report, registry, types).

Closing as part of the pre-Wave-3 stabilization sweep. Opened 2026-08-21; `main` has since gone from FR-042 to FR-092, and this branch's FR-043 id now belongs to `FR-043-quality-benchmark.md`.

PR #188, opened 2026-08-21
feat(evidence): compare generic measurements

kreneskyp:
## Summary
- query stored measurements by plan, definition, subject, scope, and source revision
- compare complete compatible populations through one measure-neutral path
- distinguish missing baseline, empty population, partial collection, and incompatible identity before arithmetic
- expose time-ordered trends with definition/provenance discontinuities
- apply ratchets only from caller-supplied direction and tolerance, and report whether the baseline was actually applied
- render deterministic canonical JSON and review-oriented Markdown

## Validation
- make test: 48 files, 551 tests passed
- focused comparison coverage: 100% statements/branches/functions/lines
- Quire validation passed with existing corpus advisories
- build completed; existing unrelated declaration diagnostics in affirm.ts, record.ts, and sbom.ts remain

## Stack
This PR is intentionally based on feat/generic-measurement-record / #187. Once #187 merges, the base can be retargeted to main.

## Scope
No analyzer names, known measure ids, universal thresholds, cross-unit normalization, or statistical significance claims are introduced.


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-044, which is already taken:**

- `spec/functional/FR-044-generic-measurement-comparisons.md`

This PR is stacked on `feat/generic-measurement-record` (#187), which mints a colliding FR-043, so both ids in the stack need renumbering together. Note also that quoin#275 edits main's `FR-044-plan-governed-measurements.md` — the requirement this branch would overwrite the id of.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `3c89ab5e` (draft, stacked on #187)

Reviewed this branch's own delta — `src/evidence/comparison.ts`, the `listMeasurementPaths` addition, FR-044 and the tests. Gates were **not** executed here — findings are from reading. The FR-044 and TC-283..289 id collisions are in my earlier comment and are not repeated.

The state machine is the strongest thing in this stack. `empty-population`, `missing-baseline`, `partial-collection`, `incompatible` and `comparable` are genuinely distinct machine states rather than one nullable result, compatibility is checked before arithmetic so a changed unit can never surface as a zero delta, and `ratchet: { requested, baselineApplied }` records that a policy was asked for separately from whether it was applied. That last pair is exactly the distinction the program keeps finding missing elsewhere.

**Verdict: FAIL** — one compatibility axis is absent from both the code and the requirement, and it is the axis most likely to produce a false regression.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | Tool identity is never a compatibility check, so a tool or config change is compared as a numeric delta | `src/evidence/comparison.ts` (`pairIncompatibilities`), FR-044-AC-3 |
| FND-002 | medium | The missing-subjects branch reports `samplingCountMismatches: []` without computing them | `src/evidence/comparison.ts` (`compareMeasurementSets`) |
| FND-003 | medium | `environment.attributes` is carried by the record and read by nothing | `src/evidence/comparison.ts` (`pairIncompatibilities`) |
| FND-004 | low | `regression` aggregates a `boolean \| null` with `.some()`, which coerces unknown to false | `src/evidence/comparison.ts` |
| FND-005 | low | Markdown table cells hold raw JSON keys, so a `\|` in a subject or repository id breaks the table | `renderMeasurementComparisonMarkdown` |

### FND-001 — the one axis that is not checked is the one that moves the number

`pairIncompatibilities` compares five things: `plan.id`, `plan.definitionVersion`, `unit`, `environment.id`, `sampling?.id`. The word `tool` does not appear in `comparison.ts`, and `MeasurementIncompatibility["reason"]` has no `tool` or `configuration` member.

`MeasurementRecord` requires `tool: { name, version, configurationDigest }`. Those three fields exist so that an observation can be attributed to the thing that produced it. Nothing reads them.

Failure scenario: per-function complexity at `r1` measured by an analyzer at 8.0.0, and at `r2` by 9.0.0, where 9.0.0 changed how it counts ternaries. Both records exist — #187's record identity includes `sourceRevision`, so they occupy different paths and are both retained. `compareMeasurementSets` finds one key on each side, all five checked axes agree, and it returns `state: "comparable"` with a delta of +4. Under any `increase-is-regression` policy that is a regression, attributed to the code. It is a tool upgrade.

The same holds for `configurationDigest` — a changed rule set for the same analyzer version moves every value and is reported as a code change.

This is a requirement gap first: FR-044-AC-3 enumerates "definition, unit, environment, or sampling identity" and omits tool, so the code faithfully implements a criterion that is missing an axis. Worth being explicit that this does not conflict with AC-2 — "without branching on measure or tool name" forbids *special-casing per tool*, which is right; comparing tool identity for compatibility is the opposite thing and is what makes measure-neutrality safe rather than blind.

Note also how this compounds with #187: that record's path identity excludes tool and environment, so the store cannot hold two tool versions for one revision, and this comparator will not notice when it has them across revisions. The same blind spot at both layers, and the second one turns it into a false regression report rather than a silent overwrite.

Suggest adding `tool-version` and `tool-configuration` to the `reason` union and to AC-3.

### FND-002 — a field reported as empty that was never computed

```ts
if (missingFromBaseline.length > 0 || missingFromCandidate.length > 0) {
  return {
    state: "partial-collection",
    missingFromBaseline,
    missingFromCandidate,
    samplingCountMismatches: [],      // not computed, reported as none
    ratchet: { requested, baselineApplied: false },
  };
}
```

The sampling-count comparison happens further down and is reachable only when both populations have identical key sets. So when the populations differ at all, `samplingCountMismatches` is emitted as `[]` — indistinguishable, in the JSON and in the Markdown ("Sampling-count mismatches: none"), from a population where they were compared and agreed.

Failure scenario: baseline covers subjects {A, B}, candidate covers {A, C}, and A's `sampleCount` is 1000 in the baseline and 10 in the candidate. The result reports the two missing subjects correctly and states that there are no sampling-count mismatches. A reviewer resolves the missing subjects, re-runs, and only then discovers that A's numbers were never comparable either — one round trip later than necessary, having been told otherwise.

This is the same shape as the recurring finding across this program: a value that reads as a clean result when the check did not run. The keys present on both sides are a well-defined set even in the partial case, so computing the mismatches over the intersection and reporting both costs nothing. If the intent really is "not evaluated in this state", then the field wants to be `null` rather than `[]`, and the renderer wants to say "not evaluated".

### FND-003 — attributes that describe the environment and gate nothing

`environment` is compared by `environment.id` alone. `environment.attributes` — an open `Record<string, string>` that the record schema explicitly admits — is never read.

`environment.id` is caller-authored, so two genuinely different machines can share `"ci"` while their attributes record `{cpu: "m1"}` and `{cpu: "xeon"}`. For a latency measure that is the whole story, and the comparison reports `comparable`.

Either compare attributes when both sides declare them and emit an `environment` incompatibility on a difference, or state in FR-044 that `environment.id` is the sole comparable identity and attributes are descriptive only. Right now a reader of the record type will reasonably assume attributes participate.

### FND-004 — a nullable aggregated by truthiness

```ts
regression: policy ? deltas.some((delta) => delta.regression) : null,
```

`MeasurementDelta.regression` is typed `boolean | null`. `.some()` coerces, so a `null` — "could not be judged" — contributes as `false`, and an aggregate of one unknown and no positives reports `regression: false`.

This is unreachable today: `deltaFor` assigns a boolean on every path when `policy` is present, and the zero-baseline case is intercepted as an incompatibility before any delta is built. So it is a latent mismatch between the type and the aggregation rather than a live defect — but the type is what a future change reads. `deltas.some((d) => d.regression === true)` plus an explicit `null` when any delta is `null` keeps the three-valued contract the rest of this file is careful about.

### FND-005 — JSON blobs in a Markdown table

`measurementKey` returns `JSON.stringify({scope: {...}, subject: {...}})`, and the renderer drops that string into a table cell inside backticks. Two consequences: a repository, path or subject id containing `|` splits the row, and every row of a "deterministic review-oriented output" carries a ninety-character JSON blob in its first column.

Deterministic it is. Readable it is not, and the pipe case is a real break rather than a cosmetic one. A rendered `kind:id @ repository/path` with the raw key kept in the JSON output would serve both audiences.

### Verified good

- Compatibility strictly precedes arithmetic, and the doc comment says why ("it can never be mistaken for a zero delta or a clean result"). Every early return carries `baselineApplied: false`, so a caller cannot read a ratchet outcome from a comparison that did not happen.
- `relativeDelta: baseline.value === 0 ? null : ...` with the type comment "`null` when the baseline is zero, where a ratio has no meaning" — and separately, a relative *policy* over a zero baseline is refused as an incompatibility rather than silently producing Infinity. Two different correct answers for two different questions about the same zero.
- `validatePolicy` refuses a non-finite or negative tolerance up front, so a `NaN` tolerance cannot make every comparison pass.
- `mixed-revision` and `duplicate-subject` are checked per side and named per side (`<baseline>` / `<candidate>`), so the reader knows which population is malformed.
- `measurementTrend` marks discontinuities per point and keeps `start` distinct from `continuous`, so the first point is not silently claimed as comparable to nothing.
- `compareStrings` is a bare `<`/`>` comparison rather than `localeCompare`, which is what CON-4 asks for and is the kind of thing that is usually got wrong.
- All seven TC ids appear in describe titles, so the matrix row binds to real tests — same as #187, and still the thing #190 is missing.

kreneskyp (commented):
## Correction to my review above — #191 already fixes FND-001 and FND-003

Having now read the tip of this stack, two of my findings are resolved there rather than outstanding. Recording it here so the review above is not read as work still to do.

- **FND-001 (tool identity never a compatibility axis)** — quoin#191 adds `"tool"` and `"configuration"` to `MeasurementIncompatibility["reason"]` and compares `name@version` and `configurationDigest` in `pairIncompatibilities`.
- **FND-003 (`environment.attributes` read by nothing)** — quoin#191 changes the environment comparison from `baseline.environment.id` to `canonicalJson(baseline.environment)`, so attributes participate.

Both are real against this branch as it stands, so the verdict holds if #188 is ever landed or reviewed on its own — but the remedy is #191, not a separate change here.

**FND-002 stands.** The `partial-collection` branch still returns `samplingCountMismatches: []` without computing them, in #191's tree as well as this one.

Same correction applies to FND-001 on #187 (record identity omitting tool/environment/sampling): #191 replaces the per-record path with an atomic `MeasurementCollection` whose identity hashes `{definitionVersion, repository, sourceRevision, tool, environment, sampling}`, which closes it.

kreneskyp:
Superseded by `main`. Measurement comparison landed as `src/evidence/adapters/differential-report.ts` within the adapter registry.

Closing as part of the pre-Wave-3 stabilization sweep. Opened 2026-08-21; this branch's FR-044 id now belongs to `FR-044-plan-governed-measurements.md` on `main`.

PR #189, opened 2026-08-21
test(perf): enforce bundle-scale budget

### LINKED ISSUE #133: [Quoin#133] NFR-011: no performance measurement exists; add a bundle-scale budget
Stated by **NFR-011** (spec-ahead-of-code) and not yet implemented.

## Problem

quoin has **no performance measurement of any kind** — no benchmark, no timing assertion, no threshold anywhere in the repository. Verified by sweeping `package.json` scripts and the whole test tree.

That was defensible while quoin was scaffolding run by hand. It is not now: `quoin evidence audit` and `quoin completeness` are merge gates, and a gate people wait on is a gate people route around.

## Scope

- A generated bundle of 250 documents.
- Each bundle-scale command walked against it: `completeness`, `advise`, `evidence audit`, `matrix`.
- Assert wall-clock against NFR-011's threshold (target < 2s, threshold < 5s).
- Assert the **pass count** by counting reads, not by inferring from timing — a regression from one pass to two must fail even on hardware fast enough to stay inside the budget.

## Why a budget rather than a prediction

The number exists so that a change making a bundle walk an order of magnitude slower fails a test instead of being absorbed as "CI got slower". 5s over 250 documents is roughly twice the current cost of the most expensive command: headroom without licensing a regression.

It also fixes the cost model in writing — these commands are O(documents) with a single pass, and a change making one O(documents × obligations) is a design change, not a tuning question.

## Non-goals

- Measuring `quire` subprocess time; the engine's own NFRs govern that.
- Optimisation. This ticket is the measurement.

kreneskyp:
## Summary
- close NFR-011's spec-ahead-of-code gap with a generated 250-document/obligation performance test
- assert the owned completeness phase stays below 5 seconds and performs exactly one pass / 250 document read attempts
- assert the owned advisor and auditor cores each process 250 derived obligations/evidence records below 5 seconds
- correct the requirement boundary: Quire subprocess time is excluded, and matrix is delegated to ix-flow rather than a Quoin bundle walk
- expose an optional bundle-read observer so pass count is measured directly instead of inferred from wall time

## Validation
- make test: 47 files, 525 tests passed
- focused performance test: 4/4 passed
- Quire validation passed with existing corpus advisories
- build completed; existing unrelated declaration diagnostics in affirm.ts, record.ts, and sbom.ts remain

Closes #133


kreneskyp (commented):
## Code review + gap analysis — head `2a15797f` (draft)

Reviewed the diff, the new tests, NFR-011 and the matrix rows. Gates were **not** executed here — findings are from reading, plus the `git show origin/main` checks quoted below.

Clean on the identity check that blocks the other seven drafts: this PR edits `NFR-011`, it mints no new id.

Two things in here are better than the PR title suggests. `test(perf)` undersells a real crash fix, and the `BundleReadObserver` counting assertion is the strongest gate in the file. The findings are about the parts of the same test that are measured by wall clock and by grep.

**Verdict: CONDITIONAL** — the flakiness risk is the one that matters, because this repo's local suite is the only gate.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | Wall-clock thresholds and a timing *ratio* run in the default lane, on a suite that is the repo's only gate | `tests/bundle-scale-performance.test.ts:192,213,216,236,239` |
| FND-002 | medium | The delegation boundary is proved by grepping one source file for strings | `tests/bundle-scale-performance.test.ts:250-258` |
| FND-003 | medium | A crash fix ships under a `test(perf)` title with no mention in the PR body | `src/auditor/audit.ts:247` |
| FND-004 | low | `observeBundleRead` has no production caller, so the budget is measurable only from a test | `src/completeness/run.ts` |

### FND-001 — a timing ratio is the flakiest assertion shape there is

```ts
expect(measured.elapsedMs).toBeLessThan(THRESHOLD_MS);        // 250 docs, 5s
expect(measured.elapsedMs).toBeLessThan(CORE_THRESHOLD_MS);   // each 10k phase, 1s
expect(elapsed.at(-1)! / elapsed.at(-2)!).toBeLessThanOrEqual(3);
```

The absolute thresholds are the usual hazard: a shared runner under parallel load turns a 5s budget into a coin flip, and the failure says "slow", which is indistinguishable from the regression the budget exists to catch.

The ratio is worse. It divides two `performance.now()` measurements, so noise in *either* compounds. At these sizes the 5k phase is short enough that scheduler jitter is a meaningful fraction of it — a 5k run that happens to land fast inflates the ratio past 3x while the 10k run is entirely healthy. The assertion fails most often precisely when the machine is behaving well on one sample and not the other, which is the opposite of a useful signal.

This matters more here than in a repo with a separate perf lane. CI is deliberately off in this program and the local suite is the only gate, so an intermittently red test is not an annoyance to be retried — it is either a merge blocked for no reason or, worse, a suite people learn to re-run until it passes, which is how a genuine regression gets waved through.

The fix is already half-written in this same file. Split the lane:

- **Default lane** — keep the observer counts. `expect(events.filter((e) => e.kind === "pass")).toHaveLength(1)` and the 250-document read count are deterministic, they cannot flake, and they are what NFR-011-M-2 actually states ("one pass per invocation"). A second pass reintroduced by a refactor fails this instantly and unambiguously.
- **Named benchmark lane** — move every `elapsedMs` assertion and the growth ratio behind an explicit tag so they run when someone is measuring, not on every commit.

That keeps the regression detection and drops the coin flip. It would also let the matrix row stop claiming coverage that depends on how loaded the machine was.

Related: the matrix row for NFR-011 flips from "Spec-ahead-of-code" straight to "Covered" on the strength of these timings. With the split above, the honest row is "covered by the pass/read-count invariant; timings tracked in the benchmark lane" — which is both true and stable.

### FND-002 — the boundary is asserted by grep

```ts
expect(matrix).toContain("extends FlowCommand");
expect(matrix).toContain("await this.launch(flags)");
expect(matrix).not.toMatch(/readBundle|runQuire/);
...
expect(readFileSync(path, "utf8")).toContain("runQuire(");
```

The claim being tested is real and worth testing — "`matrix` is excluded because Quoin delegates it directly to ix-flow, so timing it here would attribute another component's work to Quoin". But reading the command's source text is the weakest way to establish it.

Two concrete failures. Rename `FlowCommand` in a refactor and this goes red with nothing behavioural changed. Move a `readBundle` call into a helper the command imports and the regex, which reads only that one file, sees nothing — the exclusion silently stops being true while the test stays green.

Source inspection is a legitimate last resort for properties no runtime path can reach, but this one is reachable: instantiate the command with a stubbed launcher and assert it never touched the bundle reader — the `BundleReadObserver` this PR just added is exactly the instrument for it. Zero observer events from the `matrix` path is a behavioural proof of the same claim, and it survives a rename.

### FND-003 — a crash fix is riding along unannounced

```diff
-      const behind = bindings.filter(
+      const behind = runBindings.filter(
         (b, i) => runs[i].commit !== input.headCommit,
       );
```

This is not a performance change. `runs` is built from `runBindings`, so when any binding is scan-backed, `bindings.length > runs.length` and `runs[i]` is `undefined` for the tail — `runs[i].commit` throws `TypeError: Cannot read properties of undefined (reading 'commit')`. Concretely: one obligation with two bindings, one scan-backed and one run-backed, with `input.headCommit` set, and `audit()` dies.

Credit where due — the comment thirty lines below already documented that exact hazard for the `vacuous` computation, and you found the second instance of it and did tag the regression (`// TC-176`, with the matrix row extended to say "including vacuity and HEAD freshness"). That is the right work.

The finding is only that it is invisible from outside: the title is `test(perf): enforce bundle-scale budget` and the body says nothing about it. A reader deciding whether this PR is safe to defer sees a performance change; a reader bisecting a future `audit()` crash will not find it. Worth a line in the body and, ideally, its own commit.

The same applies, less sharply, to the `methodConformance` rewrite — replacing `runs[bindings.indexOf(b)]` with carried `{binding, run}` pairs removes a real `indexOf`-on-objects hazard, not just an O(n^2).

### FND-004 — a measurement seam with no production caller

`observeBundleRead` threads a `BundleReadObserver` from `AssessOptions` through `assessBundle` into `readBundleFrontmatter`. This is the right pattern — an injected observer that measures rather than a `if (testing)` branch that bypasses — and it passes the seam check cleanly.

But nothing in `src/commands/` passes it. So the budget NFR-011-M-2 states is observable only from a test, and a second pass introduced in production is caught only if someone happens to run this suite. If the intent is that the budget holds in the field, wiring the observer to a counter behind a verbose or diagnostic flag would make it so; if the intent is test-only, saying that in the doc comment stops the next reader from wondering.

### Verified good

- Dropping `schemaVersion: STORE_SCHEMA_VERSION` from the two `writeBindings` call sites is a correct no-op: `writeBindings` takes `Pick<BindingsFile, "bindings">` and sets `schemaVersion` itself (`src/evidence/store.ts:477` on main). Two call sites can no longer disagree with the store about its own version.
- `indexCatalog` preserves the old `filter` semantics exactly — deduping `id`/`class` per method via a `Set` means a method whose id equals its class is still returned once, and per-key insertion order is catalog order, so no output changes. The comment on the `entriesBySuite` dedup makes the same promise explicit for malformed legacy input ("the first reported outcome remains authoritative"). Both are the kind of note that stops a later reader from "simplifying" the behaviour away.
- `evidenceFor` now derives its runs from the obligation's own suites rather than passing the whole run list, which is a correctness improvement as much as a speed one.
- The observer emits `pass` and `document` as distinct event kinds rather than a bare counter, so the test can assert one pass *and* 250 reads separately. That distinction is what makes it a real gate.

kreneskyp:
Superseded by `main`. `spec/non-functional/NFR-011-bundle-scale-performance.md` is already on `main` carrying the same statement this PR proposes — 5 seconds on a 250-document bundle, from a measurement rather than an estimate, scoped to `quoin completeness`, `quoin advise` and `quoin evidence audit`.

Closing as part of the pre-Wave-3 stabilization sweep. Opened 2026-08-21.

Issue #180, opened 2026-08-21
[Quoin#180] vite build prints TS diagnostics in evidence modules (schemaVersion Pick mismatch, RunEntry predicate)


kreneskyp:
Found during the WP2 dependency unbreak (not introduced by it — pre-existing on main): `vite build` prints non-fatal TS diagnostics in `src/commands/evidence/affirm.ts`, `src/evidence/adapters/sbom.ts`, `src/evidence/record.ts` — `schemaVersion` not in `Pick<BindingsFile,"bindings">`; `RunEntry` type-predicate mismatches. Build exits 0, so nothing gates on them. Fix the types (adjacent to #138's RunEntry.score discriminator work — consider fixing together).

PR #190, opened 2026-08-21
feat(assurance): add bidirectional graph analyses

### LINKED ISSUE #152: Graph-analysis views (fan-out, change-impact closure, churn) were scoped in #79 and never specced
Split out of agent-ix/quoin#79, whose Scope section reads:

> Graph analyses (fan-out, change-impact closure, churn) are **in-scope for the spec as cheap views**, even if implementation lands later.

Implementation landed later and the spec never did. Verified 2026-08-19:

- no FR mentions them — `grep -niE "fan-out|change-impact|churn" spec/` finds only unrelated prose in `spec/review.md`
- no matrix row, no AC, no ticket
- `dist/commands/` ships `advise assurance completeness matrix review to-plan update write` — no graph view

So the promise exists only in a closed ticket's body, which is the least discoverable place it could be.

## Why they are still worth building

The evidence store already holds the graph these views would read — bindings keyed on `(obligation, suite)`, runs per commit, and the obligation records `quire coverage` mints. Nothing new needs collecting.

- **fan-out** — how many obligations a single suite discharges. A suite carrying fifty is a single point of failure the ratchet cannot see.
- **change-impact closure** — given a changed requirement, which suites and which obligations become suspect. The store already computes suspicion per binding; this is the transitive view of it.
- **churn** — which obligations have been re-affirmed most often. A criterion re-affirmed five times is one whose wording is unstable, which is a spec finding rather than an evidence finding.

## Not scheduled before the battletesting cycle

Deliberate. New feature surface would itself need battletesting, and that cycle should start against a fixed set. Filing so the gap is tracked rather than invisible.

## Scope, when it starts

- An FR declaring the three views and what each reads from the store.
- Read-only, like `quoin assurance` (FR-040) — collects nothing, writes nothing, and where the graph cannot answer, says so rather than rendering a partial view as complete.
- Deterministic output, so a diff means the evidence moved.

## Related

- agent-ix/quoin#79 — where they were scoped
- FR-030 — the store they would read
- FR-040 — the precedent for a read-only view over it

kreneskyp:
Closes #152.

## What changed

- specify and ship `quoin graph` with fan-out, change-impact, and churn views
- normalize declared relationship semantics into typed impact directions rather than walking an undirected component
- connect documents to current obligations, production-code `implements` scope, and evidence bindings
- trace requirements down to scoped source symbols and changed paths/`path#symbol` ids back to requirements, obligations, suites, and upstream claims
- keep implementation scope distinct from verification evidence
- label suite-sharing exposure separately and deduplicate obligation-level affirmations copied across suite bindings
- keep unreadable documents, unresolved/external targets, unsupported relationship semantics, orphan obligations/implementations/bindings, and unknown query ids explicit as incomplete results
- prevent full external `ix://` targets from aliasing local ids with the same final segment
- add deterministic JSON/Markdown output and a real command-path test

## Verification

- full suite: 48 files, 532 tests passed
- focused graph suite: 10 tests passed
- ESLint and Prettier passed
- Quire spec validation passed (existing repository warnings only; no new FR-045 warning)
- production build completed; it still prints the pre-existing TypeScript diagnostics tracked by #180 in `affirm.ts`, `record.ts`, and `sbom.ts`
- real requirement-origin pilot: changing FR-030 reached 11 downstream documents, 5 upstream context documents, and explicitly reported unresolved cross-repository/loose-reference limits
- real code-origin pilot: `tests/quire-contract.test.ts` resolved through Quire implementation scope to FR-001, its four obligations, and upstream FR-026/StR-001

Engineering-Assurance-Plan: EA-007


kreneskyp (commented):
## Code review + gap analysis — head `1b2c4b09` (draft)

Reviewed `src/graph-analysis/`, the command, the tests, FR-045 and the matrix rows. Gates were **not** executed here — findings are from reading, plus the GitHub API checks quoted below.

This is the implementation of quoin#152, and the design does the hard part right: direction is declared per verb rather than inferred, an undeclared verb becomes a stated limitation instead of a guess, and `sharedSuiteExposure` is kept separate from `suspectObligations` rather than promoted. The findings below are about the completeness signal and the traceability, not the graph maths.

**Verdict: FAIL** — one blocking identity collision, one finding that makes the honesty flag stop discriminating.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | FR-045 is already taken on `main` by a different requirement | `spec/functional/FR-045-bidirectional-trace-analysis.md` |
| FND-002 | high | `complete` is a whole-graph flag stamped on every view, so it is false for reasons the view cannot be affected by | `src/graph-analysis/analysis.ts` (`common`) |
| FND-003 | medium | Nine of ten declared TC ids bind to no test; TC-300 appears nowhere | `tests/graph-analysis.test.ts`, `spec/matrix.md` |
| FND-004 | medium | `analyzeChurn` reads raw bindings, so a retired obligation can top the ranking | `src/graph-analysis/analysis.ts` (`analyzeChurn`) |
| FND-005 | medium | `bundleIdentity()` takes the first document with the fields, not the master-requirements document | `src/graph-analysis/analysis.ts` (`bundleIdentity`) |

### FND-001 — FR-045 collides with main

This branch's merge-base is `2026-08-21T20:27Z`. `origin/main` minted `spec/functional/FR-045-portfolio-measurement-report.md` on 2026-08-27 (`0f23155`, "feat(qa): govern root assurance portfolios (#250)"), and now runs through FR-049.

Merging puts two different requirements under FR-045. Worse, quoin#191 — stacked three deep on #187 — also mints an FR-045 (`FR-045-measurement-intake-and-code-health-adapters.md`), so there is a three-way claim on the same id from `main`, this PR, and a sibling draft. The same collision affects #187/#194 (FR-043), #188 (FR-044), #192 (FR-046), #193 (FR-047) and the non-draft #275 (FR-046/047/048).

Renumber past main's highest before anything else here matters, and re-run the matrix.

### FND-002 — an honesty flag that is always false says nothing

```ts
function common<T extends object>(graph: TraceGraph, analysis: T) {
  return { ...analysis, complete: graph.complete, limitations: graph.limitations };
}
```

with

```ts
graph.complete = graph.limitations.length === 0;
```

`analyzeFanOut` reads exactly one thing: `graph.obligationSuites`. It cannot be affected by an `unsupported-relationship`, an `unresolved-relationship`, a `duplicate-document-id`, or an `orphan-implementation`. But any one of those sets `graph.complete = false`, and fan-out reports `complete: false` with a limitations list that has nothing to do with its rows.

Failure scenario: a bundle whose documents use one relationship verb outside the fourteen in `TARGET_TO_SOURCE`/`SOURCE_TO_TARGET` — which is the expected case, since quoin's edge vocabulary is open and this pair of sets is a closed allow-list. One such edge anywhere makes fan-out and churn permanently incomplete, for every bundle, forever. An operator learns within a week that `complete: false` on this view means nothing, and stops reading it.

That matters more here than usual because it is the specific property quoin#152 asked for: "where the graph cannot answer, says so rather than rendering a partial view as complete". A flag that says so unconditionally is the same as one that never says so.

`analyzeChangeImpact` already shows the fix — it refines rather than inherits:

```ts
complete: graph.complete && unknown.size === 0,
```

Push that downward. Fan-out is affected by `orphan-binding` and `unreadable-document`; churn by `orphan-obligation` and `unreadable-document`; the relationship limitations belong to change-impact alone. Each view then carries the limitations that could actually have altered its own answer, and `complete: false` regains its meaning.

### FND-003 — ten declared rows, one bound test

`spec/matrix.md` declares TC-291 through TC-300. In `tests/graph-analysis.test.ts` the only TC markers are:

```
line 7   /** FR-045 — bidirectional trace graph analyses (TC-291..TC-298). */
line 360 /** `quoin graph` reachable command path (FR-045-AC-9, TC-299). */
line 422 describe("TC-299 graph command integrates the authored and evidence graphs", ...)
```

Two things follow.

**TC-300 appears nowhere in this PR.** Not in the header range, not in a describe, not in a comment. Its matrix row is marked complete and cites FR-045-AC-10 (the changed-production-path walk). The test that actually covers that behaviour is `it("walks from a changed production path back to requirements and evidence")`, sitting untagged inside the change-impact describe. So the row is green against a test id that does not exist in the tree.

**The other nine bind only to a file, not to a test.** This repo's own convention is a header range *plus* a per-test marker — `tests/assurance.test.ts` on main carries `FR-040 — the assurance-case view (TC-221..TC-230, ...)` in the header and then `// TC-261` / `// TC-262` on the individual tests. Only TC-299 follows it here.

Failure scenario for the whole class: delete `it("counts distinct obligation-to-suite edges without imposing a threshold")`. Every matrix row stays green and the header range still claims TC-293. Nothing can tell you which row lost its test. Add `// TC-29N` above each `it`, and give TC-300 to the production-path test.

### FND-004 — churn can rank a retired obligation first

```ts
export function analyzeChurn(graph: TraceGraph, bindings: Binding[]): ChurnAnalysis {
  for (const binding of bindings) {
    for (const affirmation of binding.affirmations ?? []) { ... events.set(binding.obligation, byIdentity); }
```

`buildTraceGraph` already computes `obligationIds` and emits an `orphan-binding` limitation for every binding whose obligation is not current. `analyzeChurn` ignores both and reads the raw `bindings` array.

Failure scenario: an obligation is removed from the spec but its bindings and eight affirmation events remain in the store. Churn ranks by `affirmationCount` descending, so it lands at the top of the report. Per this PR's own framing, a high-churn obligation "is a spec finding rather than an evidence finding" — so the report sends the author to re-word a requirement that no longer exists, and the `orphan-binding` limitation that would have explained it is buried in a list attributed to all three views (FND-002).

Either filter to `graph.obligations`, or keep the row and mark it, but do not rank a retired obligation against live ones on the same axis.

### FND-005 — the authority is whichever document is first

```ts
/** The master-requirements document declares the local ix URI authority. */
function bundleIdentity(documents: BundleDocument[]): BundleIdentity | null {
  for (const document of documents) {
    const org = document.frontmatter.org;
    const component = document.frontmatter.component ?? document.frontmatter.name;
    if (typeof org === "string" && org.length > 0 && typeof component === "string" && component.length > 0)
      return { org, component };
  }
  return null;
}
```

The comment says master-requirements; the code says "the first document in `input.documents` order that has `org` and either `component` or `name`". `name` is a common frontmatter key — profiles, plans and module docs carry it.

Failure scenario: a bundle where some document with `org: agent-ix, name: finding-quality` sorts ahead of `spec/spec.md`. `identity` becomes `agent-ix/finding-quality`. `localTarget` then returns `null` for every `ix://agent-ix/quoin/FR-NNN` target, so **every** authored `ix://` edge becomes an `unresolved-relationship` and `documentEdges` collapses to whatever bare-id relationships exist. Change-impact reports an empty downstream closure with `complete: false` — which, given FND-002, is what it reports anyway. The graph is silently empty and the signal that would say so is already saturated.

Select the document by its declared type (or require `component`, not `component ?? name`), and treat "no master-requirements document" as a named limitation rather than `null`.

### Verified good

- Splitting the verbs into `TARGET_TO_SOURCE` / `SOURCE_TO_TARGET` with an explicit `unsupported-relationship` for anything else is the right call — the alternative, inferring direction from the verb string, is how a change-impact view starts reporting confident nonsense.
- `localTarget`'s refusal to take the last path segment of an external `ix://` URI, with the comment saying why (`other/FR-001` would alias to local `FR-001`), is a real aliasing bug avoided rather than a stylistic note.
- `analyzeChangeImpact` keeps `sharedSuiteExposure` out of `suspectObligations` — exposure is reported, not promoted, which is the distinction that makes the view usable rather than alarming.
- `unique(...)` keyed on NUL-joined tuples and every output run through `sorted`/`compare` gives genuine byte-determinism, which AC-8 needs.
- `analyzeChurn` deduplicating by `(commit, who, note)` rather than counting suite bindings is correct, and the comment explains why `affirm()` makes that necessary.

kreneskyp:
Superseded by `main`, which carries a **superset** of this work. Verified before closing:

- `src/graph-analysis/` on `main` has `analysis.ts`, `index.ts`, `input.ts`, `load.ts`, `render.ts` — this branch proposes three of those five.
- `src/commands/graph/` on `main` has `change-impact.ts`, `churn.ts`, `common.ts`, `fan-out.ts`, `index.ts`.
- Tests on `main`: `graph-adapters`, `graph-analysis`, `graph-command`, `graph-portfolio-command`, `graph-portfolio` — five files against this branch's two.

Closing as part of the pre-Wave-3 stabilization sweep. Opened 2026-08-21; this branch's FR-045 id now belongs to `FR-045-portfolio-measurement-report.md`.

PR #191, opened 2026-08-21
feat(evidence): add code-health measurement intake

kreneskyp:
## Summary
- add a producer-neutral evidence measure command that transcribes completed observations into generic MeasurementRecord records
- add pure adapters for Rust/Python/general structural metrics, separate churn and age, and clone-pair observations
- add a dependency-cruiser finding adapter with explicit anti-vacuity checks
- fail closed on incomplete, empty, duplicate, outside-repository, or population-mismatched collections

## Design boundary
This depends on #188. Producers run outside Quoin; authored plans own definitions and policy. The engine does not embed universal thresholds, grades, composite scores, or cross-tool normalization.

## Verification
- make test (49 files, 562 tests)
- make lint
- quire validate spec/functional/FR-045-measurement-intake-and-code-health-adapters.md
- built CLI help smoke test


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-043, FR-044 and FR-045, which are already taken:**

- `spec/functional/FR-043-generic-measurement-record.md`
- `spec/functional/FR-044-generic-measurement-comparisons.md`
- `spec/functional/FR-045-measurement-intake-and-code-health-adapters.md`

This is the deepest stack in the repo — `main` -> #187 -> #188 -> this PR — so it carries every collision below it plus its own. FR-045 in particular is claimed four ways: `main`'s portfolio-measurement-report, quoin#190's bidirectional-trace-analysis, and this branch's measurement-intake adapters.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `804427b4` (draft, tip of `main -> #187 -> #188 -> this`)

Reviewed this branch's own delta only. Gates were **not** executed here — findings are from reading. The FR-043/044/045 and TC-290..296 collisions are in my earlier comment and are not repeated.

### First: this PR already fixes two of the findings I filed downstack

Worth stating plainly before anything else, because it changes what to do with the parent PRs rather than with this one.

- **#188 FND-001 — tool identity never a compatibility axis.** Fixed here. `MeasurementIncompatibility["reason"]` gains `"tool"` and `"configuration"`, and `pairIncompatibilities` compares `name@version` and `configurationDigest`.
- **#188 FND-003 — `environment.attributes` read by nothing.** Fixed here. Environment comparison moves from `baseline.environment.id` to `canonicalJson(baseline.environment)`, so attributes participate.
- **#187 FND-001 — record identity omits tool/environment/sampling, so a re-measure overwrites.** Fixed here. `measurementCollectionPathFromValidated` hashes `{definitionVersion, repository, sourceRevision, tool, environment, sampling}`, so two tool versions or two environments at one revision no longer collide.

So those three are real against #187 and #188 as they stand and neither can merge alone, but the remedy is this branch rather than a separate change. If the stack is ever split or partially landed, that is the thing to watch.

`writeCanonicalAtomic` — same-directory temp plus `renameSync`, with the temp removed on failure — is a genuine durability improvement over the plain write it replaces, and moving from per-record files to one atomic collection is what makes it possible.

**Verdict: CONDITIONAL** — no defect in the collection model. The findings are in the adapters, where two of them disagree about what an unnamed symbol means.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | An unnamed symbol is silently dropped by one adapter and aborts the run in another; neither records a limitation | `src/evidence/adapters/measurement-observations.ts` |
| FND-002 | medium | `value` is a maximum chosen by the adapter, and `distribution` is compared by nothing | `fileDistributions`, `src/evidence/comparison.ts` |
| FND-003 | medium | `radonAdapter` merges two same-named blocks when `lineno` is absent, understating `count` and skewing `mean` | `radonAdapter` |
| FND-004 | low | jscpd subject ids join paths with unescaped `\|` and `#` | `jscpdAdapter` |

### FND-001 — the same situation, three different answers

`MeasurementAdapterResult.limitations` is documented as "Anything omitted or uncertain." Only `normalizedMeasurementAdapter` can ever populate it — every other adapter returns `unique(...)`, which hardcodes `limitations: []`. And `measure.ts` treats a limitation as fatal:

```ts
if (result.limitations.length > 0) {
  this.error(`measurement collection is incomplete: ${result.limitations.join("; ")}`, { exit: 2 });
}
```

So the channel is a refuse-channel, not a record-channel, and the six tool adapters cannot use it. What they do instead is inconsistent:

**`rustCodeAnalysisAdapter` silently skips.**

```ts
if (space.kind === "function" && name !== "" && name !== "<anonymous>") {
  ... addValue(values, path, ...)
}
```

An anonymous closure contributes nothing. The file's `distribution.count` is short by one, `mean` is computed over the wrong denominator, and `maximum` is wrong if the closure was the worst. Nothing anywhere says a symbol was omitted.

**`lizardAdapter` throws.**

```ts
if (name === "" || name === "(anonymous)") {
  throw new AdapterError(this.name, `CSV row ${index + 1} has no stable named symbol`);
}
```

Lizard emits `(anonymous)` for lambdas as a matter of course, so the first lambda in the repository aborts the whole measurement run.

Failure scenario, one command each on a mixed repo: the Rust run succeeds and quietly reports understated complexity for every file containing a closure; the Python run exits non-zero on the first lambda and records nothing. Neither says "N unnamed symbols were omitted", which is exactly what `limitations` was built for and what `--expected-count` would then reconcile against.

The design already has the right answer in it. Have the adapters push a limitation and omit the row, and let `measure.ts` decide — the current fail-closed behaviour is a reasonable default, but a `--accept-limitations` escape (recorded into the collection, not discarded) turns "unusable on real code" into "usable with the omission written down". At minimum the two adapters should agree with each other.

### FND-002 — the adapter picks the statistic, and nothing reads the rest

```ts
value: distribution.maximum,
unit: "control-flow-path-count",
distribution: { count, minimum, maximum, mean },
```

`value` is the primary comparable number, and `fileDistributions` sets it to the worst function in the file. `comparison.ts` compares `.value` and nothing else — `distribution` does not appear in it.

Two consequences.

**The distribution is write-only.** #187 schematised `count`, `minimum`, `maximum`, `mean`, `standardDeviation` and `quantiles`, this branch computes four of them per file, the collection persists them, and no code path ever reads them back. Concretely: file A has function complexities `[3,3,3,3,20]`, file B has `[19,19,19,19,20]`. Both report `value: 20`, both compare as a zero delta, and the four numbers that distinguish them are carried in the record and ignored.

**Choosing max over mean is a policy decision made in the engine.** FR-044-CON-1 says "Quoin SHALL NOT supply a default direction, tolerance, or universal threshold", and the same reasoning applies to a summary statistic: a max-based measure regresses when one new complex function is added and stays flat when twenty functions are refactored around an untouched worst case. That is a defensible measurement policy and it is not declared anywhere — the plan and definition version are supposed to carry that meaning.

Either name the statistic in the adapter name (`...-max-cyclomatic-per-file`) so the definition version pins it, or emit the statistic as part of the record so a consumer can tell which one it is looking at. And if `distribution` is meant to be comparable, `pairIncompatibilities`/`deltaFor` need to say so.

### FND-003 — a dedup key that falls back to a wildcard

```ts
const line = typeof block.lineno === "number" ? String(block.lineno) : "?";
const identity = `${path}#${owner}::${name}@${line}`;
if (!seenBlocks.has(identity)) { addValue(...); seenBlocks.add(identity); }
```

The dedup exists because radon reports a method both in `blocks` and under a class's `methods`. But when `lineno` is absent the key degenerates to a constant `"?"`, so two genuinely distinct blocks with the same owner and name collapse to one.

Failure scenario: radon output without `lineno` containing two same-named functions (an overload pattern, or a conditional redefinition). One is counted. The file's `distribution.count` is 1 short, `mean` is computed over the wrong denominator, and `maximum` is wrong if the dropped block was the larger. Silently — `addValue` is simply not called.

A missing `lineno` means the record cannot be deduplicated safely. Treating that as an `AdapterError`, or as a limitation under FND-001, is honest; a wildcard that merges is not.

### FND-004 — unescaped separators in a composite id

```ts
const pair = [left, right].sort();
id: `${pair[0]}|${pair[1]}#${fragmentDigest}`
```

`|` and `#` are legal in POSIX paths. A file named `a#b.ts` or `a|b.ts` produces an id that cannot be unambiguously split back into its parts, and two different pairs can in principle render to the same string. The digest makes accidental collision unlikely, but the id is the record's stable subject identity — worth percent-encoding the two separators or using a JSON-encoded composite the way `measurementKey` does in `comparison.ts`.

### Smaller note

`normalizedMeasurementAdapter` validates `subject`, `path`, `unit` and `value` field by field, then passes `distribution` through as `item.distribution as MeasurementDistribution` on the strength of an is-an-object check. The collection schema catches a malformed distribution before write, so nothing bad is stored — but the error the user sees is a JSON-pointer schema failure rather than an adapter error naming the offending observation. Given every sibling field gets a located message, this one is worth the same.

### Verified good

- `if (result.observations.length === 0)` refusing with *"record an explicit aggregate zero rather than treating empty output as clean"* is the single best line in this stack. A missing parser or an empty traversal cannot become a clean baseline, which is precisely the failure this program keeps documenting elsewhere.
- `--expected-count` is a caller-declared reconciliation against silent truncation — the adapter cannot quietly return half the file set and have it accepted.
- `normalizedPath` normalises separators *before* testing for escape, so `..\foo` on Windows is caught by the same check as `../foo`, and it re-tests `isAbsolute` after `relative`.
- `csvRows` is a real CSV reader — escaped quotes, CRLF, and an explicit throw on an unterminated quoted field — rather than a `split(",")`.
- `unique()` refusing a duplicate stable subject means an adapter cannot emit two observations that would silently collapse to one in the store.
- `writeCanonicalAtomic` cleans up its temp file on failure and renames within the same directory, so a crash mid-write leaves either the old collection or the new one.

kreneskyp:
Superseded by `main`. Code-health measurement intake landed as the adapter registry under `src/evidence/adapters/` (10 adapters incl. `cargo-mutants.ts`, `sarif.ts`, `junit.ts`, `sbom.ts`) feeding `recordRun()` in `src/evidence/record.ts`.

Closing as part of the pre-Wave-3 stabilization sweep. Opened 2026-08-21; this branch also depends on #187 and #188, both closed as superseded for the same reason.

PR #192, opened 2026-08-21
feat(evidence): record producer trust decisions

kreneskyp:
## Summary
- add a strict, use-specific evidence-producer trust decision record
- compare accepted and observed producer/configuration/adapter/corpus/contract/environment contexts
- report changed selected triggers as invalidated and absent observation as unobserved
- render producer reliance as assurance context without changing claim status

## Boundary
This is a small control-plane JSON record in Quoin's existing evidence store. It does not run, certify, approve, or globally qualify tools, and it supplies no analyzer allowlist.

## Verification
- make test (47 files, 529 tests)
- make lint
- quire validate spec/functional/FR-046-evidence-producer-trust.md
- built CLI help smoke test


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-046, which is already taken:**

- `spec/functional/FR-046-evidence-producer-trust.md`

FR-046 is claimed three ways: `main`'s `FR-046-record-semantic-data-planes.md`, this PR, and the non-draft quoin#275's `FR-046-clause-discharge.md`. #275 is further along, so it will likely win the id — plan the renumber around that.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `9baa95f8` (draft)

Reviewed `src/evidence/trust.ts`, the store and render additions, FR-046 and the tests. Gates were **not** executed here — findings are from reading, plus the `git show origin/main` checks quoted below. The FR-046 id collision is in my earlier comment and is not repeated.

The core idea is right and it is the idea the epic asks for: trust is scoped to a *use*, not stamped on a tool, and absence of an observation is reported as `unobserved` rather than folded into either acceptance or rejection. `assessTrust` returning `status: "unobserved"` with an empty `triggeredBy` is precisely the "distinguish could-not-check from checked-and-clean" property flagged on quire-rs#386.

**Verdict: FAIL** — one fail-open in the invalidation mechanism itself, plus a second identity collision the branch cannot see.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | An adapter appearing only in the observed context is never compared, so reliance survives it | `src/evidence/trust.ts` (`superRefine`, `differs`) |
| FND-002 | high | TC-297..TC-300 are also minted by quoin#190, off the same base | `spec/matrix.md`, `tests/trust-decision.test.ts` |
| FND-003 | medium | `readTrustDecisions` reports skipped records through an optional out-parameter that defaults to a throwaway | `src/evidence/store.ts` |
| FND-004 | medium | `permittedDecisions` bounds reliance and nothing enforces the bound | `src/evidence/trust.ts`, FR-046 |
| FND-005 | low | `decidedAt` accepts anything `Date.parse` accepts — a third timestamp rigour level in one repo | `src/evidence/trust.ts` |

### FND-001 — the trigger set is fixed against the accepted context only

```ts
if (decision.acceptedContext.adapter && !uniqueTriggers.has("adapter")) {
  context.addIssue({ ... message: "must include adapter when an adapter is relied upon" });
}
```

`adapter` is the one trigger left out of `REQUIRED_TRIGGERS`, and it is required only when the *accepted* context already names an adapter. Nothing requires it when the accepted context has none.

Failure scenario: a producer is trusted reading its native output — `acceptedContext.adapter` is absent, so `revalidateOn` legitimately omits `"adapter"`. Later the pipeline routes that producer through an adapter, and `observedContext.adapter` appears. `assessTrust` filters `decision.revalidateOn`, which does not contain `"adapter"`, so `differs("adapter", ...)` is never called. The decision reports `accepted`. A new, unreviewed translation layer now sits between the producer and the claim, and the mechanism whose entire job is to notice context change reports no change.

The asymmetry is the bug: `superRefine` guards the direction where the adapter is already known and leaves the direction where one is introduced. Either make `adapter` unconditionally required in `REQUIRED_TRIGGERS`, or have `assessTrust` treat "a context field present in `observedContext` but absent from `acceptedContext`" as a trigger regardless of the selected set — an unaccepted field appearing is exactly the event a reliance decision should not survive.

While you are there: `differs("adapter", ...)` compares via `JSON.stringify`. Both sides come from `parsed.data`, so Zod normalises key order to schema order and it is correct today. It is still a structural comparison done by string, which stops being safe the moment `adapter` gains an optional field or a nested object. A field-by-field comparison costs two lines and does not depend on that.

### FND-002 — four test-case ids claimed twice

`origin/main`'s matrix occupies TC-270 through TC-276; TC-277 is the first free id in that block. This branch and quoin#190 both fork from `main` and both allocate forward from there without knowing about each other:

- **this PR** — `describe("TC-297 ...")` through `describe("TC-302 ...")`
- **quoin#190** — matrix rows TC-291 through TC-300

TC-297, TC-298, TC-299 and TC-300 name completely different tests in the two branches. Here TC-297 is "a decision is scoped to one use"; in #190 it is churn deduplication. Whichever merges second silently redefines four ids that the first one's matrix already points at, and the matrix keeps rendering as complete because every row still has *a* test.

This is the same failure as the FR-046/047/048 collision I raised earlier, one level down, and it is not confined to this pair — the 277..302 range is contested across #187, #188, #190, #191, #192 and #194, all forked from the same 2026-08-21 base. A single allocation pass across the open branches is worth doing before any of them merge, and the same "lint the merged result, not the branch" fix covers both id families.

### FND-003 — the skip channel defaults to a black hole

```ts
export function readTrustDecisions(repo: string, skipped: string[] = []): TrustDecision[] {
```

FR-046-AC-6 requires that Quoin "reports unreadable or invalid stored decisions without hiding the remaining decisions". The mechanism is an optional mutable array the caller has to remember to pass. `readTrustDecisions(repo)` — the shorter, more natural call — collects the skips into a throwaway and discards them.

Failure scenario: a hand-edited or half-written `ETD-004.json` fails `validateTrustDecision`. A caller using the one-argument form gets a clean-looking list of the decisions that parsed, with nothing to indicate that a reliance decision exists and could not be read. For this record type that is the worst direction to fail in: a trust decision that cannot be parsed is not the same as no trust decision, and the shorter call cannot tell them apart.

`{ decisions, skipped }` as a return value makes it impossible to ignore, and costs the callers one destructure. The out-parameter shape puts the burden on every future caller to opt into honesty.

### FND-004 — a bound with no enforcement

`use.permittedDecisions` is `z.array(z.string().min(1)).min(1)` — an open vocabulary, deduplicated by `superRefine`, sorted by `assessTrust`, rendered nowhere in `renderCase`, and compared against nothing. It is the field that says what this producer's output may be used to decide, and no code path checks that an actual use is in it.

That is a gap rather than a defect: FR-046-AC-1 requires Quoin to *identify* permitted downstream decisions, and it does. But a reliance bound that is recorded and never enforced will read to an auditor as a control, and it is a comment. Either state explicitly in FR-046 that enforcement is out of scope and belongs to the consuming layer, or add the criterion that a consumer asserting a decision outside `permittedDecisions` is reported.

### FND-005 — three timestamp contracts in one repo

```ts
decidedAt: z.string().refine((value) => !Number.isNaN(Date.parse(value)), {
  message: "must be an ISO-8601 timestamp",
}),
```

The message says ISO-8601; the check says "whatever `Date.parse` accepts". `Date.parse("12/25/2026")` returns a number, so a locale-ambiguous US-format date validates as an ISO-8601 timestamp and is stored as one.

Same repo, same concept, three rigour levels right now: quoin#275's `instant()` regex-pins the ISO shape *then* calls `Date.parse`; quoin#187's `collectedAt` is regex-pinned to UTC-with-`Z`; this is `Date.parse` alone. #187's pattern is the one to copy.

The same fragmentation has appeared in the digest patterns across this stack — `^sha256:[0-9a-f]{64}$` in #275, `^[A-Za-z0-9][A-Za-z0-9._+-]*:[A-Fa-f0-9]+$` in #187, `^[a-z0-9]+:[a-f0-9]+$` here. Three definitions of "a digest" in one evidence store. Worth one shared `$defs`/Zod primitive rather than three.

### Verified good

- `status: "unobserved"` with `triggeredBy: []` for an accepted decision with no observed context is the right answer and the one the program keeps asking for. It is not `accepted`, not `invalidated`, and a consumer can act on the difference.
- `triggeredBy` names *every* changed trigger rather than short-circuiting on the first, so the report says what to revalidate rather than only that something changed.
- Scoping the decision to `use.id` rather than to the producer means two uses of one tool can carry different reliance — AC-5 — which is what stops this becoming a global tool badge, and the TC-297 test name says so explicitly.
- `.strict()` on every Zod object, including nested `adapter` and `validationEvidence` items, plus the `refine` requiring digest-or-reference on each evidence entry. Same posture as #187's schema.
- The render block prefixes the section with "These decisions are context for the case; they do not make a claim supported" and `renderCase` does not touch claim status — AC-7 held in the code, not just asserted in a test.
- Every TC id appears in a describe title, so the matrix binds to real tests. That is the thing #190 is missing.

PR #193, opened 2026-08-22
feat(evidence): evaluate profile-selected independence

kreneskyp:
## Summary

- record normalized evidence lineage without parsing assurance-profile Markdown in Quoin
- evaluate profile-selected independence across actor, implementation/toolchain, technique, data source, and review path
- surface insufficient separation as an audit finding and render assessments as assurance context
- preserve audit and assurance JSON shapes when no independence policy is selected

## Verification

- 538 tests passed
- build passed
- Quire validation passed (pre-existing warnings only)

## Stack

Drafted on top of feat/evidence-producer-trust / PR #192 because independence relies on the evidence provenance model introduced there.


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-047, which is already taken:**

- `spec/functional/FR-047-profile-selected-evidence-independence.md`

FR-047 is claimed three ways: `main`'s `FR-047-allocate-semantic-module-ownership.md`, this PR, and quoin#275's `FR-047-authored-assurance-arguments.md`. This PR is also stacked on `feat/evidence-producer-trust` (#192), whose FR-046 collides the same way, so both need renumbering together.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `db594712` (draft, stacked on #192)

Reviewed this branch's own delta — `src/evidence/independence.ts`, the auditor wiring, the command surface, FR-047 and the tests. Gates were **not** executed here — findings are from reading. The FR-047 id collision is in my earlier comment and is not repeated.

The lineage model is well chosen: five named dimensions, all optional, a policy that must *select* which ones matter for a given obligation, and an explicit refusal to infer independence from roles or vendor names (CON-1). `requireKnownPolicyObligations` throwing on a stale obligation id — "refuse a stale/typoed projection rather than silently evaluating nothing" — is the exact opposite of the failure class this program keeps hitting, and it is good to see it written deliberately.

**Verdict: FAIL** — one finding suppresses the rest of the audit for an obligation, and the assessment cannot say it had nothing to assess.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | An `insufficient-independence` finding `continue`s past every other check for that obligation | `src/auditor/audit.ts` |
| FND-002 | high | `status` is binary, so "no lineage recorded" is reported identically to "genuinely common-mode" | `src/evidence/independence.ts` (`assessIndependence`), FR-047-AC-4 |
| FND-003 | medium | TC-305, TC-306, TC-307 and TC-309 bind to a describe-block range, not to a test | `tests/independence.test.ts`, `spec/matrix.md` |
| FND-004 | low | `assessIndependence` never checks that the bindings belong to the requirement's obligation | `src/evidence/independence.ts` |

### FND-001 — the audit stops at the first independence finding

```ts
if (assessment.status === "insufficient") {
  findings.push({
    kind: "insufficient-independence",
    obligation: obligation.id,
    severity: "medium",
    summary: assessment.summary,
  });
  continue;
}
```

That `continue` leaves the per-obligation loop. Everything after it — the vacuity check, the HEAD-freshness check, method conformance, the remaining discharge accounting — never runs for that obligation.

Failure scenario: an obligation is selected by the profile, its two suites share an actor, one of its bindings names a symbol the run never reported, and both runs are behind HEAD. The audit emits exactly one finding: `insufficient-independence`, at medium. The reader fixes the lineage, re-runs, and two further findings appear that were true the whole time. Worse in the other direction — a team that decides the independence requirement is not worth satisfying now has an obligation whose vacuity and freshness are permanently unaudited, and nothing says so.

This file already argues the opposite position, sixty lines above, about a different pair of findings:

> an obligation with an uncatalogued method is BOTH undischarged AND unknown-method, and hiding either behind the other cost a reader the ability to see both facts.

Independence is orthogonal to discharge in exactly the same way. Push the finding and fall through, as the `unknownMethodFinding` call does.

If the `continue` is deliberate — the reasoning being that an obligation without independent evidence should not be assessed on other axes — that needs to be stated as a criterion, because none of AC-1 through AC-8 says it and the resulting report cannot be distinguished from "we checked the other axes and they were clean."

### FND-002 — the assessment cannot report that it had nothing to assess

`IndependenceAssessment.status` is `"satisfied" | "insufficient"`. The dimension detail carries `missingSuites`, and the `summary` prose says "missing on suite-a, suite-b" — but the machine-readable status collapses two very different situations:

1. Two suites recorded full lineage and genuinely share a toolchain. That is an assurance finding.
2. No binding has ever recorded lineage at all. That is a data-collection gap; nothing was assessed.

Both produce `status: "insufficient"` and both produce an `insufficient-independence` finding at medium severity. A consumer gating on `status` — or reading the findings list, which carries only the summary string — cannot separate them.

Failure scenario: a repo adopts an AssuranceProfile with independence requirements before any `evidence record` run supplies lineage. Every selected obligation reports `insufficient`. The report is identical in shape to a mature repo whose evidence is real and genuinely common-mode. The first needs someone to start recording lineage; the second needs a second independent evidence line. The report recommends neither, because it cannot tell which it is looking at.

This is the property engineering-assurance#3 names directly — "Distinguish unavailable, not-computed, and not-applicable observations" — and the same shape as `complete` on quoin#190 and `open` on quoin#275. A third status (`unevaluated`, when every selected dimension is missing on every binding) plus a distinct finding kind would carry it, and the data to compute it is already in `dimensions[].missingSuites`.

FR-047-AC-4 is worth rereading against this: "Shared and missing lineage values remain visible and prevent a false independent result." The values do remain visible in `dimensions`. What AC-4 does not require — and what is missing — is that the two causes stay distinguishable in the *status*.

### FND-003 — four ids share two describe blocks

`spec/matrix.md` claims TC-303..TC-309. The test file has three describes:

```
describe("TC-303 strict lineage and policy boundaries", ...)      // 2 its
describe("TC-304..TC-307 relationship independence", ...)         // 4 its
describe("TC-308 lineage persistence and TC-309 assurance context", ...)  // 3 its
```

So TC-304 through TC-307 name one block of four tests with no per-id marker, and TC-308/TC-309 share a block of three. Which `it` discharges TC-306 ("with no independence policy, the auditor preserves its existing output") is inferable from the names — `it("does nothing when no profile requests independence")` is clearly it — but nothing binds them, and the matrix row cites all seven ids as covered.

Failure scenario: delete `it("keeps common actor and missing parser lineage visible")`. TC-305's row stays green because its describe block still exists and still passes. This is milder than #190, where nine of ten ids bound to nothing at all, and stricter than it needs to be to fix: a `// TC-305` comment above each `it` matches what `tests/assurance.test.ts` does on main.

### FND-004 — an unenforced precondition

`assessIndependence(profile, requirement, bindings)` never checks that each binding's `obligation` equals `requirement.obligation`. Both call sites in `audit.ts` are correct — the loop passes `bindings` for the current obligation, and the sweep passes `bindingsByObligation.get(requirement.obligation) ?? []` — so there is no live defect.

It is worth noting because the failure mode is silent and favourable: pass the full binding list and two suites bound to *different* obligations satisfy the requirement, reporting independence that was never demonstrated for the obligation in question. One `filter` inside the function, or a thrown error, makes the contract enforceable rather than remembered.

### Verified good

- `requireKnownPolicyObligations` refuses rather than evaluating an empty set. A policy naming a renamed or deleted obligation would otherwise assess nothing, report nothing, and read as clean — the precise failure this program has been finding elsewhere, and it is pre-empted here.
- The doc comment on `assessIndependence` states the rule that makes the whole thing honest: "A dimension absent on either side cannot demonstrate separation." Two suites sharing one parser stay visibly common-mode rather than passing on a missing value.
- The pairwise search requires `first !== undefined && second !== undefined && first !== second` on **every** selected dimension, so partial separation never satisfies a multi-dimension requirement.
- `a.suite !== b.suite` means one evidence relationship cannot satisfy both sides — AC-3 — which is the obvious way this kind of check gets fooled.
- Both call sites and the sort are deterministic: `ordered` is suite-sorted before the scan, so `satisfiedBy` names the same pair on every run.
- `valueFor` re-trims and re-checks emptiness even though `EvidenceLineageSchema` already does, which matters because bindings come off the store rather than through that schema.
- `report.independence` is attached only when a policy exists, so CON-2's "no policy, no change" holds in the output shape and not merely in the values.

PR #194, opened 2026-08-22
feat(workflows): integrate assurance context

kreneskyp:
## Summary

- make `/specify` discover and carry applicable Assurance Profile, Impact Tier, Tool Trust Plan, and Measurement Plan context
- make direct and ix-flow-backed `/spec-review` select the same applicable assurance artifacts and preserve the declared analysis-set invariant
- reject truncated `all` review sets and add cross-harness eval scenarios for assured and unassured repositories

## Stack

This draft is based on `fix/spec-review-ears-parity` / #186 because the workflow contract relies on the corrected EARS review-set parity. It also expects the architecture-evaluation schema from `spec-artifacts-process` #58 before release.

## Boundary

- skills orchestrate and select context; they do not duplicate Quire parsing or assurance rubrics
- profile applicability never silently makes every assurance artifact mandatory
- ix-flow retains explicit human gates

## Verification

- 528 tests pass
- build and lint pass
- Quire spec validation passes
- deterministic eval fixtures TC-EV-058 and TC-EV-059 are committed

## Draft blockers

- release of the architecture-evaluation schema dependency
- live Codex and Claude runs of the disposable eval fixtures after explicit authorization to transmit those fixture/worktree instructions to external model endpoints

The small evidence TypeScript fixes in the first commit repair baseline type errors exposed by the full build: callers no longer pass the store-owned schema version, and SBOM maps state their nullable projection type explicitly.


kreneskyp (commented):

|---|---|---|

main now runs through FR-049.

**This branch adds FR-043, which is already taken:**

- `spec/functional/FR-043-assurance-aware-spec-workflows.md`

quoin#187 also mints an FR-043 (`FR-043-generic-measurement-record.md`), so this id is claimed three ways: `main`, this PR, and a sibling draft. This PR is also stacked on `fix/spec-review-ears-parity` (#186), which mints no new ids and is clean on this check.

### What to do

kreneskyp (commented):
## Code review + gap analysis — head `1248efc6` (draft, stacked on #186)

Reviewed this branch's own delta — the intake invariant, the workflow definition, the skill changes, FR-043 and the tests. Gates were **not** executed here — findings are from reading, plus the greps quoted below. The FR-043 id collision is in my earlier comment.

`reviewSelectionConsistent` is a good gate and it is wired where it belongs — `intake -> analyses_run`, so a malformed selection is refused before any analysis runs rather than after. The three-mode profile model (`none` / `recommend` / `require`) with recommendations explicitly advisory, and the comment saying why only required selections are validated against the installed schema, is the right shape.

**Verdict: FAIL** — eight matrix rows bind to nothing, and a workflow contract change ships without a version bump.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | All eight TC ids this PR mints appear in no test; `grep -nE 'TC-[0-9]+'` over the three test files returns nothing | `spec/matrix.md`, `tests/*.test.ts` |
| FND-002 | high | TC-277..TC-284 are already minted by #187 and #188 for different tests | `spec/matrix.md` |
| FND-003 | medium | `def.yaml` gains three interview questions and a gate with no version bump | `workflows/review/def.yaml` |
| FND-004 | medium | In `require` mode the final gate passes trivially when `profile_analyses` is empty | `scripts/invariants.js` (`selectedAnalysesCovered`) |

### FND-001 — eight green rows, zero bound tests

`spec/matrix.md` adds TC-277 through TC-284, each marked complete, each citing an AC. Running the repo's own convention over the three test files this PR touches:

```
$ grep -nE 'TC-[0-9]+' tests/eval-assert.test.ts tests/spec-review-vocab-drift.test.ts tests/specify-assurance.test.ts
(no output)
```

No `// TC-NNN` comment, no id in a `describe`, no id in an `it`. The describes are `"Codex rollout command assertions"` and `"assurance-aware specify contract"`; the tests are named by behaviour only.

The tests themselves are real — `it("cannot truncate all or bypass a required profile selection")` drives the actual invariant through a dynamic import and asserts on structured codes, which is the right way to test a gate. The problem is purely that nothing connects them to the eight rows claiming they exist.

Failure scenario: delete `it("keeps a recommended profile advisory when the user chooses base")`. TC-279's row stays complete, TC-279 still cites FR-043-AC-3, and no coverage figure moves. There is no query — by grep, by tooling, or by eye — that identifies which of the eight rows lost its test.

This is the strongest instance of a pattern across this stack. #187, #188 and #192 put one TC id in each describe title and are fine. #193 groups four ids into one describe, which is weak but traceable. #190 has a header range covering nine ids and one bound test. This PR has none. `tests/assurance.test.ts` on main shows the house style: a header range plus `// TC-261` above the specific test.

Note also that #186 — this PR's own base — added `tests/spec-review-vocab-drift.test.ts` untraced, and this PR extends that same file, still untraced. Worth fixing both together since the file is shared.

### FND-002 — the id block is already taken twice over

`origin/main` occupies TC-270..TC-276. This PR allocates TC-277..TC-284 from there. So do two sibling branches, forked from the same 2026-08-21 base:

- **#187** — TC-277..TC-282 (`describe("TC-277 one measure-neutral record shape covers unlike observations")` and five more)
- **#188** — TC-283..TC-289

Every one of the eight ids here names a different test in a different branch. TC-277 is "`/specify` names all three installed assurance artifact types" in this PR and "one measure-neutral record shape covers unlike observations" in #187.

Because of FND-001 the collision is invisible on this side: with no id in any test, a merge produces two matrix rows for TC-277 and there is nothing to compare them against. Renumber past #193's TC-309, and do it after FND-001 so the ids land somewhere a tool can see them.

### FND-003 — a changed workflow contract under an unchanged version

`def.yaml` gains three interview questions (`assurance_profile`, `profile_selection_mode`, `profile_analyses`) and a new invariant on the `intake -> analyses_run` transition. The `version:` field is untouched — `grep -n "^[+-]version:"` over this diff returns nothing.

Your own base commit sets the precedent in the other direction: #186 bumped `0.2.0 -> 0.2.1` for a strictly smaller change (three prose strings and one entry in `ALL_ANALYSES`).

Failure scenario: an instance persisted mid-run under the pre-profile 0.2.1 resumes against the post-profile 0.2.1. Its `operation_request` has no `profile_selection_mode`, so `norm(undefined || "none")` yields `"none"` and it takes the non-profile path — which happens to be correct. But nothing in the instance records which contract it started under, and the next change to these fields will not have that luck. Two definitions that answer differently to the same interview must not share a version string.

Bump to `0.3.0` — the interview grew required-adjacent structure and a transition gained an invariant, which is a minor-version change, not a patch.

### FND-004 — the final gate has no floor of its own

```ts
const selectedAnalysesCovered = ({ instance }) => {
  ...
  if (profileMode === "require") {
    selected = normalizedList(request?.profile_analyses);
  } else if (reviewSet === "all") {
    selected = [...ALL_ANALYSES];
  }
  if (selected.length === 0) return true;
```

The comment above this function calls it the "Hard check (final gate)". Before this change its `all` branch expanded to `ALL_ANALYSES`, a module constant that cannot be empty. Now the `require` branch draws from `request.profile_analyses`, which is caller-supplied and optional in the interview (`required: false`). An empty or absent list makes `selected` empty and the gate returns `true` — accepting a run with no SpecReview documents at all, in the mode that is supposed to be the strictest.

`reviewSelectionConsistent` does refuse `profileAnalyses.length === 0` in require mode, at intake, so the hole is closed today by a *different* invariant on an *earlier* transition. That is worth being deliberate about rather than incidental: the final gate is the one whose comment claims it is the hard check, and it currently depends on an upstream gate having run and on the request not having been amended since. A three-line floor — in `require` mode, an empty `profile_analyses` is a failure rather than a pass — makes it true on its own terms.

Same class as the `all_set_incomplete` check you added: a gate that can be satisfied by supplying less input.

### Verified good

- `reviewSelectionConsistent` is registered on the `intake -> analyses_run` transition rather than exported and forgotten, and the tests assert both the wiring (`expect(...).toContain("review_selection_consistent")`) and the behaviour. That is the pairing that stops a gate from existing only on paper.
- Every refusal returns a structured `{ ok: false, code, details }` — `unsupported_selected_analysis`, `required_profile_review_set`, `required_profile_selection_mismatch`, `all_set_incomplete`, `subset_is_empty` — so a caller can act on the reason rather than parse a message. The tests assert on the codes.
- Dropping the old `reviewSet === "all" && selected.length === 0` condition makes `all` *always* expand to the canonical vocabulary, so a truncated caller-supplied list can no longer weaken the final gate even if intake were bypassed. AC-5 says exactly that, and the change is in the strict direction.
- Keeping recommendations out of the `unsupported` check, with the comment explaining that only selections that will actually run need schema support, is a real distinction rather than an oversight — a profile may legitimately recommend an analysis this installation does not have.
- CON-2 — "the workflow SHALL NOT automatically acknowledge its human acceptance gate" — is stated as a Safety constraint rather than left implicit, which is the right place for it.

PR #275, opened 2026-08-29
Add explicit assurance records and human-owned sufficiency

kreneskyp:
## Outcome

Adds project-owned, source-neutral assurance workflows on top of validated ClauseSet bindings.

## Features

- complete clause discharge partitions: direct evidence, approved disposition, open, unresolved, and not binding
- authored assurance arguments with explicit, current human sufficiency decisions and visible assumptions and challenges
- immutable, content-addressed experiment and operational evidence records with closed producer provenance
- deterministic CLI rendering with no aggregate quality or compliance score
- separate provenance pins for the current consumer-contract CLI and historical measurement producers, so existing evidence is never relabelled as newly regenerated

## Dependencies

This PR consumes the CLI contract from quire-cli#72. Its build workflow selects exact CLI revision `f3baa7b33b06a6afd1ee9a102b56ea3f04aefdf9`; historical benchmark producer pins remain unchanged.

## Verification

At final rebased head `d55e24706ca39156face7844d51f905955dffeef`:

- 739 tests pass
- TypeScript checking, ESLint, and formatting pass
- all 20 tool-drift mutation cases pass
- all 20 verification-stack invariants pass
- change-assurance 0.2 completed an exact-head observation with a verified event chain and stopped at `review_ready`; its top claim remains open for the human decision

The release-readiness drift command is expected to remain on hold until a release tag exists. That is a release gate, not an implementation-test failure. Only project-authored code, schemas, documentation, and synthetic fixtures are included.


kreneskyp (commented):
## Code review + gap analysis — head `42165f75`

Reviewed the clause-discharge and assurance-record layer, the Quire consumer contract, the workflow changes and the matrix rows. Gates were **not** executed here — findings are from reading, plus four GitHub API checks quoted below.

**Verdict: FAIL** — the consumer-contract pin names a quire-cli that does not contain the command this PR consumes, and the only test on that pin cannot detect it.

| ID | Sev | Summary | Refs |
|---|---|---|---|
| FND-001 | high | `cliSourceRevision` pins a quire-cli revision with no `clauses` subcommand | `src/quire/contract.ts:34` |
| FND-002 | high | The only assertion on that pin is a 40-hex-character shape check | `tests/quire-contract.test.ts:74` |
| FND-003 | medium | Four distinct causes collapse into `state: "open"`, separated only by free text | `src/assurance/discharge.ts:130-146` |
| FND-004 | low | `stringValue` validates the trimmed value and returns the untrimmed one | `src/assurance/discharge.ts:~345` |
| FND-005 | low | `renderDischargeReport` wraps unescaped values in backticks | `src/assurance/discharge.ts:~175` |

### FND-001 — the pin and the dependency disagree

`contract.ts` declares:

```ts
/** Exact quire-cli commit exercised by CI for this consumer contract. */
cliSourceRevision: "bcface2714a958a328f3427714650ab2df71030f",
```

The PR body says the build workflow selects exact CLI revision `f3baa7b33b06a6afd1ee9a102b56ea3f04aefdf9`. Those are different commits, and the difference is the whole feature:

```
$ gh api repos/agent-ix/quire-cli/contents/src/commands/clauses.rs?ref=bcface27...
404 Not Found

$ gh api repos/agent-ix/quire-cli/compare/bcface27...f3baa7b3 --jq '.files[].filename'
... src/commands/clauses.rs ...          # added in that range
```

`bcface27` is `chore: pin reviewed Quire validation stack (#71)` — it predates quire-cli#72 and has no `clauses` verb at all. So the revision this PR declares as "exercised by CI for this consumer contract" cannot emit a `clause-binding-v1` payload; the schemas vendored alongside it describe output it cannot produce.

Either the constant was not advanced when the CLI dependency was, or the workflow builds something the constant does not name. Both are the same underlying problem — see FND-002.

### FND-002 — the pin's gate cannot fail

```ts
expect(QUIRE_CONTRACT.cliSourceRevision).toMatch(/^[0-9a-f]{40}$/);
```

Forty hex characters. `0000…0000` passes. A commit from a different repository passes — I checked, `bcface27` does not exist in quire-rs either, and nothing here would have noticed if it had been pasted from there.

This is precisely the defect class you filed against quire-rs#386 last week: *the toolchain cannot distinguish verified from could-not-verify, and reports success for both.* The neighbouring pin has a real gate — `sourceRevision` is checked against the recorded schema digests in "hashes exactly what contract.ts records". `cliSourceRevision` has a shape check standing in for a verification.

The workflow builds `.ci/quire-cli/Cargo.toml` from a recursive submodule, so the ground truth is the gitlink. A test asserting

```ts
execFileSync("git", ["rev-parse", "HEAD:.ci/quire-cli"]).toString().trim()
  === QUIRE_CONTRACT.cliSourceRevision
```

turns FND-001 into a red test rather than a claim in a PR body. Worth doing before this merges — a stale pin here is invisible by construction, and this repo is the one that is supposed to catch that shape elsewhere.

### FND-003 — `open` is doing four jobs

```ts
if (!fact) { open.push(entry(clause, "open", "no discharge fact")); continue; }
const current = currentAttestation(fact.attestation, asOf);
if (current !== null) { open.push(entry(clause, "open", current, fact)); continue; }
```

`currentAttestation` returns one of three distinct strings — `"attestation expiry is not after attestation"`, `"attestation is in the future"`, `"discharge fact is expired"` — and all three land in the same partition as "nobody has attested this at all", carrying the same machine-readable `state: "open"`.

Failure scenario: a discharge fact is submitted by a producer whose clock is a day ahead. The clause reports `open`, indistinguishable in `binding.open.length` from an unattested clause. A dashboard counting open clauses shows the correct number and the wrong reason; the operator chases missing evidence that was in fact supplied and rejected. Same for a lapsed attestation, which is an actionable renewal, not an absence.

The module doc opens by insisting applicability and discharge stay separate facts and that unresolved stays outside the denominator — the same argument applies one level down. A `reasonCode: "no_fact" | "not_yet_valid" | "expired" | "malformed_validity"` alongside the prose costs one field and makes the partition machine-readable to the depth the rest of the design already reaches.

### FND-004 — trimmed for the check, untrimmed for the value

```ts
function stringValue(name: string, value: unknown): string {
  if (typeof value !== "string" || value.trim().length === 0) throw ...
  return value;                       // untrimmed
}
```

`clauseId: " W-1"` passes validation, matches no clause in the binding report, and surfaces as `unusedFacts: [{clauseId: " W-1", reason: "unknown_clause"}]` while W-1 itself reports `open`. That is fail-closed, so no evidence is manufactured — but the diagnostic points at "unknown clause" for a clause that exists, and `unknown_clause` is the one `UnusedDischargeFact` reason a reader will assume means a typo in the id rather than whitespace around it. Returning `value.trim()` (as `parse_context` does in quire-cli) removes the case.

### FND-005 — backticks in, backticks out

`renderDischargeReport` emits ``- \`${item.clauseId}\` — ${item.force}; ...`` and ``- \`${fact.clauseId}\` (${fact.kind}): ${fact.reason}``. A clause id or a Quire-supplied reason message containing a backtick terminates the code span early. Deterministic and non-destructive, but the render is a published artifact.

### Verified good — and worth pushing upstream

`parseFact` / `parseAttestation` reject unknown fields explicitly:

```ts
function exact(value, name, fields) {
  for (const key of Object.keys(value))
    if (!expected.has(key)) throw new Error(`${name} has unknown field ${key}`);
  ...
}
```

That is the right posture, and it is exactly what quire-rs#381's clause-set input types are missing — none of `ClauseSet`, `Clause`, or `ClauseSetRights` carries `deny_unknown_fields`, so a transposed `applicabilty:` there silently makes a scoped clause bind universally. I have raised it on that PR; flagging here because this repo already demonstrates the fix.

Also good:

- `asOf` is an explicit parameter and the module states it never reads the wall clock — the discharge partition is reproducible from its inputs.
- `evidenceDigest` is pattern-checked to `sha256:<64 lowercase hex>` rather than accepted as any string.
- `unusedFacts` distinguishes `unknown_clause` / `not_binding` / `unresolved`, so a fact supplied against a clause that does not bind is visible rather than silently discarded. That is the distinction FND-003 asks you to extend to `open`.
- Duplicate facts for one clause throw rather than last-wins.

kreneskyp (commented):
## Follow-up — main moved under this branch (new since my review 20 min ago)

Not a code finding; a merge-blocking identity collision that only became true when quoin#311 landed on 2026-08-30, after this branch's merge-base (`2026-08-29T15:35Z`).

**Three requirement ids are now claimed twice.**

`spec/functional/` at this PR's head `42165f75`:

```
FR-046-clause-discharge.md
FR-047-authored-assurance-arguments.md
FR-048-append-only-assurance-evidence.md
```

`spec/functional/` on `origin/main` today, all three from `c3cbd99` "Record semantic module architecture" (merged as #311, 2026-08-30T23:39):

```
FR-046-record-semantic-data-planes.md
FR-047-allocate-semantic-module-ownership.md
FR-048-declare-authority-by-concern.md
```

Merging as-is puts two different requirements under each of FR-046, FR-047 and FR-048 — different files, different subjects, same id — plus two conflicting `spec/functional/index.md` additions.

Worth flagging *how* this got past verification rather than just that it did. The PR body reports 739 tests passing at `d55e2470`, and that is true — but it was true against a merge-base that did not yet contain #311. Nothing in the run could have seen the collision, and nothing will until after the merge, which is the same "the gate could not check, and reported success" shape as FND-002 in my earlier review. If quoin has a duplicate-id / reference-lint gate, it is worth making it run against `origin/main` merged in rather than against the branch alone; that turns "stale branch mints a taken id" into a red check on the PR.

Renumber to FR-050/051/052 (main's highest is FR-049), refresh `spec/functional/index.md` and `spec/matrix.md`, and re-run. The five findings in my earlier review still stand independently — FND-001 (`cliSourceRevision` pins a quire-cli with no `clauses` subcommand) is the one that should block on its own.

For the record, the same collision affects every open draft in this repo: all nine branch from `2026-08-21T20:27Z`, and main has since minted FR-043 (08-28), FR-044 (08-26), FR-045 (08-27) and FR-046/047/048 (08-29). #187 and #194 both mint FR-043, #188 mints FR-044, #190 and #191 both mint FR-045, #192 mints FR-046, #193 mints FR-047. Commenting on each separately.

Issue #326, opened 2026-09-01
[Quoin#326] Pinned quire contract rejects a real payload from the installed quire 0.31.0


kreneskyp:
Found while gating #322 (SR-114 FND-001). Not caused by that change: it reproduces on a stashed, unmodified tree.

## What fails

`tests/quire-contract.test.ts` TC-118 — "a real `quire coverage --json` payload validates against the pinned schema".

It builds a temporary fixture repository, runs the installed `quire coverage --scope <root> --json`, and validates the payload against the contract pinned in `src/quire/contract.ts`. `parseCoverage` returns `ok: false`.

The payload that is refused begins:

```json
{"unbacked_rows":[{"reference":"verification","document":"spec/functional/FR-001-thing.md","row_id":"FR-001-AC-1","target_ids":["FR-001-AC-1"],"line":17}],"status_lies":[],...}
```

Installed engine: `quire 0.31.0 (cli 4f6ed024, engine 0.46.0@ca7362d4)`.

## Why it matters beyond the test

TC-118 exists because the contract is what every consumer command parses coverage through. A payload the contract refuses is a payload `quoin evidence record`, `quoin report`, and the graph commands also refuse — so this is a live incompatibility with the installed engine, not only a red test.

## What is not yet established

Which side moved. The two candidates are a contract that is too narrow for a field the engine legitimately added, and an engine emitting a shape it should not. Settling it means reading the engine's current output schema against `src/quire/contract.ts`, not widening the contract until the test passes.

## Acceptance criteria

- [ ] The direction of the drift is stated as a fact: contract too narrow, or engine output wrong.
- [ ] TC-118 passes against a named, pinned quire version, and that version is recorded where the pin lives.
- [ ] If the contract is widened, the justification is independent of the test turning green.

kreneskyp:
Diagnosed. The failure scenario in the description (and in SR-114 FND-001) was too broad — **real repository exports parse fine**. Superseded by this:

## The actual refusal

```
/binding_census/0: must have required property 'tagged'
```

`quire 0.31.0 (engine 0.46.0)` emits a `binding_census` entry with no `tagged` property; the pinned contract in `src/quire/contract.ts` requires it.

## What triggers it

The census entry omits `tagged` when the repository has **no tagged symbols**. TC-118's fixture repo has exactly one Rust file whose only marker is `#[tracks("TC-999")]`, which binds nothing.

Measured against two real exports, both parse `ok: true`:

[...]

PR #325, opened 2026-09-01
Expose the producer-facing change-assurance CLI surface

kreneskyp:
Closes #322.

Quoin already implements the FR-063 change-assurance record, FR-064 proof attestation, retained-output intake, and FR-065 verification receipt. A producer had no supported way to reach any of them. This adds FR-068, `quoin change-assurance`, with seven subcommands — and nothing else: the library contracts, the stored layout, and the schema assets are unchanged.

## The surface

| Command | Does |
|---|---|
| `seal-record` | Seals an explicit record body and retains it under its digest |
| `seal-attestation` | Binds an already-produced result file to the caller's stated record, revision, proof, command, tool, configuration, environment, and result |
| `intake` | Verifies and atomically retains the exact attestation JSON and the exact output bytes |
| `receipt` | Assembles the FR-065 verification input from named inputs only and emits the receipt |
| `verify-receipt` | Re-verifies a sealed receipt's shape and digest |
| `schema` | Emits a packaged normative schema byte-for-byte |
| `recover` | Removes staging left by an interrupted intake |

## It is not a runner

No command executes the command an attestation describes, spawns a process, invokes Git, or performs a network request. TC-1327 asserts that over the sources rather than in prose, and it enforces the responsibility boundary the same way: it deletes the one permitted disclaimer sentence from each file and then requires that no claim verb survives, so any other mention of authentication, authorization, non-repudiation, or certification is a failure however it is phrased.

The only fields the surface derives are `retained_output.digest` and `retained_output.size_bytes`, read from the bytes of the file the caller names, plus the media type the caller declares. TC-1318 asserts field-by-field that everything else in the sealed attestation is the caller's, byte for byte.

`seal-record` and `seal-attestation` refuse a body that supplies `digest` (or `retained_output`). The sealing functions silently `delete` a supplied digest before hashing, so without that refusal a caller could hand in a wrong digest, get a cleanly sealed record back, and be told nothing.

## Missing evidence stays missing

`unavailable`, `not_computed`, and an absent attestation remain their own receipt outcomes and reasons; none is converted into a pass or a failure. TC-1321 drives that as a property over the three non-passing results plus the no-selection case.

Exit status is 0 for a `valid` receipt, 1 for `invalid` or `incomplete`, and 2 for a usage, parse, or integrity error. The receipt is still emitted for the first two; for the third, none is, so nothing can be mistaken for a result. An unselected stored attestation changes no receipt field (TC-1320), so a stray upload cannot discharge a proof.

## Gates

[...]

PR #328, opened 2026-09-01
Inventory campaign-native result formats and add the two missing adapters

### LINKED ISSUE #323: Inventory campaign-native result formats and close demonstrated adapter gaps
## Objective

Map the actual structured outputs produced by the eight campaign repositories to Quoin's existing adapters and add only the narrow adapters that are demonstrably missing.

## Inventory scope

JUnit-compatible test reports, SARIF, cargo-mutants, audit-script/normalized entries, Kani outputs, solver analysis reports, contract conformance JSONL, corpus reports, measurements, counterexamples, and domain-specific proof or differential reports.

## Acceptance criteria

- [ ] Every proposed adapter names a real producer, pinned sample, governed target record, and why an existing adapter cannot represent it.
- [ ] Adapters validate and transcribe; they do not execute producers or decide overall assurance sufficiency.
- [ ] Arbitrary stdout/stderr scraping is explicitly out of scope.
- [ ] Empty, malformed, partial, failed, unavailable, unsupported, and not-computed samples are covered.
- [ ] Native structured output is retained when available; console streams are retained only when they are the material diagnostic artifact.
- [ ] No adapter duplicates an Engineering Assurance or existing Quoin record family.

## Dependencies

Blocked on engineering-assurance#5. Hosted CI remains manual-only.

kreneskyp:
Closes #323.

Surveyed what the eight campaign repositories actually emit, at their `origin/main`, then added an adapter only where a real producer emits a real format no existing adapter can represent.

## The survey result

Eleven of thirteen scope items need nothing:

| Item | Verdict |
|---|---|
| JUnit, SARIF, cargo-mutants, audit scripts, advisory scans | Already covered |
| Measurements | `quoin measurement record`, not an adapter concern |
| Corpus reports | A corpus manifest is a verification *definition* Quire owns, not a result |
| Kani, solver analysis, domain proof reports | **No producer emits them.** `quire-analyze`'s entire library surface is `hello()` |
| Counterexamples | Carried inside the differential report's own cases |

Writing a reader for a format nobody produces means inventing the format, and the first real sample would then be shaped to fit the reader rather than the other way round. The inventory names the reopening condition — a real sample — instead of leaving it implied.

## The two that are real

**`contract-conformance`** reads the JSONL `quire-contract-conformance run` emits in `quire-contract-ir`. Pinned against seven unedited lines of a real 99-row run spanning all four operations and both valid and invalid fixtures; the full run's digest is recorded. Entries are keyed `<corpus>::<operation>::<fixture>` because the runner replays the same fixture id under four operations, and a bare id would let one result overwrite another.

**`differential-report`** reads `<domain>.differential-summary/v1`, produced by `tl-mltl` against R2U2 4.2-release / C2PO 4.1.0. Pinned against the committed report at `tl-mltl` `fe1c620`, unedited.

Neither existing adapter could take these: `entries` wants one object with an `entries` array, `junit` wants XML, and the finding-shaped adapters would write a conformance run into `findings/` — losing the clean-versus-unrun distinction FR-034 exists to make — or record a differential disagreement as a rule violation, which is a different claim from "two implementations disagree" and one the report does not make.

## The interesting part: a state the store cannot hold

`RunEntry.outcome` is `pass | fail | skip | error`. A differential's `unsupported` is none of them — nothing chose to omit the case, and nothing failed. Rounding it to `skip` would delete the distinction at intake, permanently and silently, for every future run of that producer.

So the adapter **does not transcribe it and names it**. `AdapterResult.unrepresented` carries the producer's own state verbatim with the reason, and `quoin evidence record` prints each one in both human and JSON output:

```
  not transcribed — closed-profile-not-mapped-v1 reported unsupported: the external
  reference does not support this case; that is neither a skip nor an error, and no
  run-entry outcome carries it
```

That is an adapter-layer field: no new record family, no stored schema change, nothing duplicated from Engineering Assurance. Extending the stored outcome vocabulary is a bigger decision than an adapter ticket, and it should be made against a count of how many real producers need it — not against this one. Recorded as SR-116 FND-002.

Both adapters fail closed on an unknown protocol or schema version, an unknown status, a missing field, and an empty result set — a run that emitted no line did not report that every fixture matched.

## A correction to PR #325

That PR reported `prettier --check .` clean. It was, when I ran it — and then I added its two review artifacts and committed without re-running it, so both landed unformatted and `prettier --check .` currently fails on `main`. Nothing about the FR-068 surface is affected. Both files are formatted here, and the correction is posted on #325. A check run before the last commit is not a check of the commit.

## Gates

- `tsc --noEmit`, `eslint`, `prettier --check .` — clean.
- `vitest run` — 926 of 927 pass; the one failure is `quire-contract` TC-118, filed as #326 and unrelated. TC-1332's pinned adapter-name list was updated, not widened: that list is deliberately the command's surface, and adding two names is a surface change the test is there to catch.
- `quire coverage --json` — FR-069 has no unbacked row and the new test file contributes no unmatched tag.

Reviews retained as SR-116 and SR-117, both CONDITIONAL.

Hosted CI remains manual-only and was not dispatched.

https://claude.ai/code/session_011EqxTnd7CWjB3NwfBdRFhj


kreneskyp:
Mergeable. All six #323 acceptance criteria are discharged: both added adapters name a real producer, a pinned real sample, a governed target record, and why nothing existing represents them; they validate and transcribe without executing anything; stdout scraping is out of scope and stated; the malformed/empty/failed/unsupported states are covered with `unsupported` named rather than collapsed; native structured output is what both read; and no record family is duplicated.

Branch protection blocks self-approval on this account, so this comment is the merge signal.

Issue #350, opened 2026-09-04
[Quoin#350] make test refuses on main: the verification stack's quire-rs lock predates quire-rs#388


kreneskyp:
**Priority:** P1
**Track:** Verification stack
**Initial board state:** Backlog.

## Objective

Unblock `make test` in this repository: its verification stack pins a `quire-rs`
revision that Track A Wave 4 has since moved past, so the gate refuses before
any test runs.

## Symptom

On a clean checkout of `main`, with a `quire-rs` checkout beside it:

```
$ make test
node scripts/verification-stack.mjs

Error: quire evidence overlay 17b80e420f719eeaf9c33f541118aa480f5d6175
contains a merge before terminal promotion from locked code
ca7362d4dacecb96f01d74d1d971327118c25917
    at assertRepository (scripts/verification-stack.mjs:268)
```

`make test` exits non-zero without running a single test.

## It is `main`, not a branch

Verified by differential: a clean clone of `origin/main` and a clone with
`spec/307-module-cookiecutter` squashed onto it fail identically, with the same
two revisions named. This is not introduced by any open pull request.

## Cause

`17b80e4` is the merge of `agent-ix/quire-rs#388` (the extraction boundary),
landed as step 4 of Track A Wave 4. The locked revision `ca7362d` predates it.
The stack's rule — that the evidence overlay must not contain a merge before
terminal promotion from the locked code — is doing exactly what it says; the
lock has simply gone stale relative to a `quire-rs` `main` this same campaign
advanced.

## Why it matters more than a stale pin usually would

The gate is placed ahead of the whole suite, so a stale lock does not degrade
one check — it removes every check in this repository at once, and it does so
by refusing rather than by passing. That is the right direction to fail in, and
it still means `make test` currently proves nothing here.

`npx vitest run` still works, which is how this was verified while the gate is
blocked; that is a workaround, not a substitute, because it skips the
provenance assertions the stack exists to make.

## Deliverables

- Re-lock the stack to a `quire-rs` revision at or after `17b80e4`, with the
  evidence overlay regenerated against it.
- State how the lock is refreshed and who refreshes it when `quire-rs` `main`
  advances, so the next campaign does not rediscover this the same way.
- Fail with a message that names the remedy — which command re-locks it —
  rather than only the two revisions.

## Acceptance criteria

- [ ] `make test` runs the suite on a clean `main` with a current `quire-rs`
      checkout beside it.
[...]

PR #351, opened 2026-09-05
Corpus measurement over the completed Wave 4 modules (#291)

### LINKED ISSUE #291: [Quoin#291] [GATE] Measure the full corpus against completed module schemas (advisory, report only)
**Priority:** P0
**Track:** Spec
**Parent epic:** https://github.com/agent-ix/quoin/issues/286
**Initial board state:** Backlog — after all Wave 4 module tickets.

## Objective

Measure every completed module schema against the governed corpus, classify failures, and publish the report. **Advisory, report only:** this campaign does not normalize or migrate corpus repositories; findings feed a later normalization campaign.

## Deliverables

- Run deterministic advisory validation over the pinned full corpus and retain raw results.
- Partition failures into contract defect, legitimate undeclared value, malformed document, missing structure, unsupported representation, stale module, and tool defect.
- Publish pass rates by module/type/repository and preserve unknown/unavailable states.
- Re-run until every failure has an owner and disposition (contract fix in this campaign, or deferred corpus fix with a named later campaign).
- Normalization batches are out of scope for this campaign.

## Acceptance criteria

- [ ] Every corpus document is validated or explicitly classified.
- [ ] No aggregate hides a weak module/type partition.
- [ ] Contract amendments cite measured legitimate cases; deferred corpus rewrites cite the accepted contract.
- [ ] All P0/P1 contract findings are resolved or explicitly accepted by a human owner.
- [ ] The final report pins corpus, module, compiler, and schema revisions.

## Dependencies and sequencing

- All module schema-completion tickets.
- agent-ix/quire-rs#385 governed corpus.
- Quire and Quoin integration tickets.

## Safety / merge gate

- [ ] New constraints remain advisory throughout this ticket.
- [ ] No corpus repository is edited by this ticket.
- [ ] A high failure rate pauses promotion; it does not justify weakening the validator.

kreneskyp:
Closes #291. **Advisory, report only** — no corpus repository was edited, nothing normalized or migrated, and no schema weakened to lower a count.

## The measurement

Population `pop:cafc3ffdf1b73a29`, corpus `ix-workspace-2026-09-04`, 40.0 s.

| | |
|---|---|
| Repositories | **251** |
| Spec documents | **7,587** |
| Modules resolved | **10** (the nine this wave merged, plus `engineering-assurance`) |
| Declared types | **90** — 61 object, 29 artifact |

**Structural conformance 7,370 / 7,501 = 98.3 %** (documents, `engine-structural-v1`)
**Properties-form census 1 / 155 = 0.6 %** (documents, `properties-form-census-v1`)

The two are never summed: different populations, different methods, and a combined figure would describe neither. 3 repositories dirty, 0 unstable, both published beside the rates.

## Every finding classified, `unknown` zero

`missing-structure` 292 · `malformed-document` 7 · `stale-module` 1 · everything else 0 · **`unknown` 0, `undispositioned` 0**.

The first run reported **300 unknown**. That was the truthful state before anything was classified, and it was published rather than papered over. Rules were then authored from sampled documents, each recording the sample that settled it — and the applier leaves an unrecognised reason as `unknown` on purpose, because a rule set that silently absorbs what it does not recognise is indistinguishable from one that got it right.

**Bad rule or bad corpus, per this repo's own rule: sampled and read.** `missing-structure` (agent-cli-daemon, agent-config-cookiecutter — the tables are genuinely absent); `malformed-document` (agent-duncan, catalog-service, cloud-manager-ui-services — column sets differ, one `Traces To` cell holds an em dash); `stale-module` (quire-rs AP-201 — carries a `concern` the pinned archetype rejects). **No rule was widened.** All are deferred to the normalization campaign with the owning repository named.

## What it could not establish

**`quire-rs#405` covers the entire population.** `IX_FILAMENT_MODULES_PATH` adds to the default install root instead of replacing it, and resolution is first-wins, so every module is reported twice and which copy answered a document is not determined by the pin. **The verdicts stand; their attribution to the pinned revisions does not**, and the report says so in its own section.

The other four ledger entries matched nothing and are reported unmatched — an entry covering nothing is fixed, mis-scoped, or was never real.

## Findings this ticket produced about its own tooling

Filed en route: **`quire-rs#405`** (module set not closed). The ledger also carries `quire-rs#402`, `quire-rs#403`, `spec-artifacts-process#81`, `quoin#347`.

Five defects were found in **my own** measurement code, every one of which reported a clean run:

1. `showAt` trimmed its output, so 84 `data_schema` digests read as mismatches. Verified one by hand before believing it — the modules were right.
2. The type parser required column-zero list items, so `engineering-assurance` contributed **zero** types and reported clean.
3. Diagnostics matched on relative paths where the engine emits absolute — every document read as `pass` with zero diagnostics.
4. The engine resolves relative globs against the process cwd, not `--scope`, so a batch validated the wrong directory and reported it clean.
5. `path` is not always a diagnostic field; the same engine emits it as a field under one invocation and only as a message prefix under another.

Each was found by falsifying against a deliberately invalid document or by a number being implausible — none by a test going red. **Reading zero of something was indistinguishable from reading nothing**, four separate times.

## Verification

- 57 tests across nine suites.
- `quire validate --scope . "spec/**/*.md"` structurally clean.
- All 13 figures in the report recomputed from the artifact and field they cite: **0 mismatched**.
- Determinism asserted over a committed fixture corpus, not this workspace — a determinism claim about one developer's machine is not reproducible by anyone else.

`make test` cannot run in this repository (**`quoin#350`**, the verification stack's `quire-rs` lock predates `17b80e4`); `npx vitest run` was used, which skips the provenance assertions the stack exists to make.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

https://claude.ai/code/session_019a1pGXRt5jL8RqVEiLvhNf


kreneskyp:
**mergeable** — verified in an isolated clone with the required sibling checkouts (`quire-rs`, `spec-objects-business`, `spec-artifacts-iso`) present.

| Gate | Result |
|---|---|
| Squash onto `origin/main`, `update-ref` | `diff:[]` `status:[]` |
| `quire validate --scope . "spec/**/*.md"` | structurally clean |
| The nine new suites, post-merge | **57 / 57** |
| Report figures recomputed from their artifacts | **13 / 13, 0 mismatched** |

## Scope of the verification, stated rather than implied

The full 1,050-test suite was killed twice by resource pressure (load 6.85, twelve node processes) rather than failing. A killed run leaves no result; a failing suite leaves a count. Reporting the first as the second is the same conflation this ticket is careful about between `could-not-run` and `fail`, so I am not doing it here.

What bounds the risk is the diff: `git diff --diff-filter=MD` shows the only **modified** files are `spec/` index and matrix documents. No existing source or test file is touched — the branch is additive to `src/measurement/` and `tests/`. The existing suite's behaviour cannot change except through the new files, and those are the 57 above.

For the record, the four failures I confirmed pre-existing on `main` earlier today — `bench-tier1` TC-975/997/998 and `quire-contract` TC-118 — are unrelated to this branch, and `make test` cannot run in this repository at all (**#350**).

## The measurement

Population `pop:cafc3ffdf1b73a29`, 251 repositories, 7,587 documents, 10 modules, 90 declared types.

- **Structural conformance 7,370 / 7,501 = 98.3 %** (documents, `engine-structural-v1`)
- **Properties-form census 1 / 155 = 0.6 %** (documents, `properties-form-census-v1`)

Never summed: different populations, different methods. 3 repositories dirty, 0 unstable, published beside the rates.

Findings: `missing-structure` 292, `malformed-document` 7, `stale-module` 1, **`unknown` 0, `undispositioned` 0**.

The first run reported **300 unknown**, and that was published rather than papered over — it was the truthful state before anything had been classified. The rules were then authored from documents that were actually opened and read, each rule recording its sample, and the applier deliberately leaves an unrecognised reason as `unknown`.

**No rule was widened.** All three classes are bad corpus, deferred to the normalization campaign with the owning repository named. This ticket measures and does not migrate, and no corpus repository was edited.

## What it could not establish

`quire-rs#405` covers the whole population: `IX_FILAMENT_MODULES_PATH` adds to the default install root rather than replacing it, and resolution is first-wins. **The verdicts stand; their attribution to the pinned contract revisions does not**, and the report carries that as its own section rather than implying a precision the run cannot support. The other four ledger entries matched nothing and are reported unmatched.

## Five defects in the measurement's own code, every one of which reported a clean run

A trimmed digest that made 84 correct schemas read as mismatches; a parser that read zero types from a module and called it clean; relative-versus-absolute path matching; the engine resolving globs against cwd rather than `--scope`; and `path` not always being a diagnostic field.

Each was caught by falsifying against a deliberately invalid document, or by a number being implausible against something already known. **None by a test going red.** Four separate times in this ticket, reading zero of something was indistinguishable from reading nothing — which is the same shape as the binder defect in `quire-rs#403` and the reason the ledger entry for it exists.

Issue #177, opened 2026-08-21
no drift gate coupling skills/spec-matrix vocabulary to the module manifest


kreneskyp:
## Symptom

PR #163 retired the ⚠️ marker from `skills/spec-matrix/SKILL.md`, both templates, and a gap-analysis reference — by hand. Nothing couples the skill's declared Status vocabulary to the module manifest (`spec-artifacts-process` `column_patterns.Status` / `traceability.status`). This is the exact drift class the whole CR-083 program addressed, now guaranteed to recur undetectably: the next time the module vocabulary changes, the skills silently teach agents the old one.

Also: `skills/spec-matrix/assets/test-matrix-example.md` rows now read `🚧 Partial` — the marker plus a note whose word is the retired concept, reintroducing the two-forms-one-meaning the PR argues against.

## Fix

- A test (or `make ci` check) that parses the vocabulary declared in `skills/spec-matrix/SKILL.md` + templates and compares it against the installed module manifest's admitted/classed sets. Fails on divergence in either direction.
- Drop "Partial" from the example rows.

Found in retro review of the trace-status-integrity batch (PR #163).

Issue #81, opened 2026-08-15
[Quoin#81] ISO 25010 completeness: N/A-acceptance verdict policy (computation moved to quire-rs#162)


kreneskyp:
Part of epic agent-ix/quire-rs#81 (ADR-0011 program). **Gated: do not start until the epic's gate clears.** P2.

**ADR-0011 placement:** L1 · computation as a quire Validator where deterministic, verdict policy in quoin (the FR-050 split). Placement detail is the first Specify question.

## Problem

Nothing checks a spec bundle for the all-functional-requirements failure mode: a corpus can be 100% AC-covered and still have no requirement anywhere for reliability, security, or maintainability. The 25010 sweep is the highest-yield completeness check and the one most often skipped.

## Scope

- Walk the ISO 25010 characteristics (functional suitability, performance efficiency, compatibility, interaction capability, reliability, security, maintainability, flexibility, safety): each must own ≥1 requirement **or** carry a written "N/A because …" record in the bundle.
- The characteristic list and the req↔characteristic mapping vocabulary are module data (spec-artifacts-iso), not engine constants — likely needs a companion change there; surface it in Specify.
- Verdict/threshold policy (hard 100% vs advisory) stays in quoin per FR-050 precedent.

## Non-goals

- Judging requirement *quality* per characteristic — that is quire-rs#83's lane. This check is purely presence/justified-absence.

kreneskyp:
**Scope note (2026-08-16):** an earlier plan draft would have added a 29119-4 technique-catalog sweep here as a sibling of the 25010 sweep. That is now subsumed: the verification-method catalog (agent-ix/quire-rs#133 + agent-ix/spec-artifacts-process#35) IS the 29119-4-shaped registry, and "every obligation's method comes from the declared catalog" is the auditor's method-conformance check (quoin#80). This ticket stays purely the 25010 characteristic sweep.

kreneskyp:
## Re-scoped to verdict policy only — 2026-08-17, P2 planning

**The computation moves to quire-rs: agent-ix/quire-rs#162.** This issue keeps the
N/A-acceptance verdict policy.

### Why — ADR-0011 placement, and the precedent it names

ADR-0011 §Placement: *needs only the spec corpus, deterministic → quire-rs core; needs the
consumer's code, test results or repo state → quoin.* "Does every declared characteristic own
≥1 requirement" is a deterministic fact about the spec alone.

[...]

PR #371, opened 2026-09-12
fix: emit and enforce one Status column

### LINKED ISSUE #370: quoin emits and accepts Coverage Status, so new repos are born with the skipped-classification defect
The TestMatrix contract collapsed `Coverage Status` and `Status` into a single `Status` column (spec-artifacts-process #87, `375fc2a`). The two names carried byte-identical vocabularies and were applied inconsistently across sibling coverage tables, so the single global `traceability.status.column: Status` selector matched only some of them and **status classification was silently skipped on the rest**.

quoin is both the generator and the validator, so it propagates the defect two ways.

## It generates it

There is no TestMatrix skeleton in any module, so the whole authoring surface is the `spec-matrix` skill's assets plus the cookiecutter:

- `skills/spec-matrix/assets/test-matrix-template.md` — lines 17, 22, 28
- `skills/spec-matrix/assets/test-matrix-example.md` — lines 17, 24
- `templates/semantic-module/{{cookiecutter.repo_name}}/spec/matrix.md` — lines 68, 77

The cookiecutter is the one that matters most: **every new semantic-module repo is born with the old header.** The template is already internally inconsistent — lines 33 and 39 say `Status` — so it teaches the split it causes.

## It accepts it

`default-modules.yaml` pins `spec-artifacts-process` at `d605caa`, which still carries 5 × `Coverage Status`. Until that moves, quoin's own validation accepts the old header and a corrected template would fail CI.

## Its own matrices carry it

`spec/matrix.md:181,1130` and `spec/evals.md:86`. Measured with the collapsed contract installed:

```
status-column-matches-nothing   2 → 0
status_lies                     9 → 43   (+34 newly visible)
```

Those 34 rows were always wrong. The check simply never ran on those two tables.

## The guard that should have caught it

`tests/skills-vocab-drift.test.ts` (TC-271) already couples the skill and its assets to the installed manifest — but gates **markers only**. Its `statusCells` helper keys on a header cell equal to `Status`, so a table headed `Coverage Status` is skipped and never checked. A stale column *name* was invisible to the very test written to stop vocabulary drift.

## Not in scope

`corpus/` is a submodule. `corpus/cases/minting/status-column-mismatch/` is a deliberate-mismatch fixture proving the engine reports the fault — the collapse makes it *more* correct. `corpus/modules/ecosystem/` is a SHA-pinned vendored copy with a documented refresh ritual, and is its own reviewed event in its own repo. Measurement and bench baselines are re-measured, never edited. No consumer repos are swept.

kreneskyp:
Closes #370.

The TestMatrix contract collapsed `Coverage Status` and `Status` into a single `Status` column (spec-artifacts-process#87, `375fc2a`). The two names carried byte-identical vocabularies and were applied inconsistently across sibling coverage tables, so the single global selector matched only some of them and **status classification was silently skipped on the rest** — after which `status_lies` was empty because nothing ran, not because the rows were honest.

quoin propagated that two ways: it **generated** the old header and it **accepted** it.

## Changes

**Root cause first** — `default-modules.yaml` pinned the module at `d605caa`, which still carries the old name. Bumped to `375fc2a`. Until this moved, validation accepted the old header and a corrected template would have failed CI.

**The generators.** No module ships a TestMatrix skeleton, so the entire authoring surface is the `spec-matrix` assets plus the cookiecutter:

| File | |
| --- | --- |
| `skills/spec-matrix/assets/test-matrix-template.md` | already inconsistent — taught `Status` on two tables, `Coverage Status` on three |
| `skills/spec-matrix/assets/test-matrix-example.md` | |
| `templates/semantic-module/{{cookiecutter.repo_name}}/spec/matrix.md` | **every new semantic-module repo was born with the defect** |

**quoin's own matrices** — `spec/matrix.md:181,1130`, `spec/evals.md:86`. Renaming makes classification run on them for the first time:

```
status-column-matches-nothing   2 → 0
status_lies                     9 → 43   (+34 newly visible)
totals / unbacked               unchanged (1029/1874, 323)
```

The 34 rows were always wrong; the check never ran. **They are not reconciled here** — no status is promoted.

**A landmine handled.** `spec/matrix.md:1135` held `⚠️ Partial` — retired by CR-031 and admitted by no vocabulary. Once the header matched it would have become a fresh validation failure. It is now `🚧` carrying the gap its own note already states ("resume/advance/gate progression is untested"), avoiding the retired concept returning as a note word per #177. The remaining ten `⚠️` cells sit in a `Coverage` column and are unaffected — verified.

## The guard that should have caught this

`tests/skills-vocab-drift.test.ts` (TC-271) already coupled the skill and assets to the installed manifest, but gated **markers only**. Its `statusCells` helper keys on a header cell equal to `Status`, so a table headed `Coverage Status` was skipped and never checked — a stale column *name* was invisible to the test written to stop vocabulary drift.

It now asserts every taught status column is named exactly as the manifest configures. **Verified red before green:**

```
AssertionError: the template names a status column `Coverage Status`,
but the manifest configures `Status`
```

## The ledger entry is updated, not removed

`src/measurement/tool-defects.ts` `status-lie-check-never-ran` (#81) stays. The contract is fixed, but the fix is **per-repository**: the check still does not run on any `tests.md` that has not renamed its header. Removing the entry would claim a measurability that does not exist for unswept repos.

## Not touched

- `corpus/` — a submodule. `corpus/cases/minting/status-column-mismatch/` is `kind: failure` with a required finding; its `Coverage Status` header is the deliberately-wrong input proving the engine reports the fault, and the collapse makes it *more* correct. Editing it would delete the regression test for quire-rs#341.
- `corpus/modules/ecosystem/` — SHA-pinned vendored manifests with a documented refresh ritual; its own reviewed event in its own repo. **Still pinned pre-collapse at `995288d`** — flagged below.
- `spec/evidence/measurements/*.json`, `bench/*.json` — re-measured, never edited.
- No consumer repos swept.

## Verification

- `make lint` clean (typecheck + eslint + prettier)
- `npx vitest run` — **105 files, 1182 tests, 0 failures**
- Cookiecutter now emits `| … | Status |` on all three tables
- `make test` (verification-stack) not run: it requires a `quire-rs` checkout at `/home/peter/dev/worktrees/quire-rs` that is absent here — an environment prerequisite, unrelated to this change

## Loose thread

`qa-corpus` still vendors `spec-artifacts-process` at `995288d`, pre-collapse, so quoin's corpus cases validate against the old two-name contract until refreshed per `VENDORED.md`. Separate repo, separate reviewed event — not done here, flagged so it is not found later as a surprise.

https://claude.ai/code/session_01J4qKWaqT3xCXMbbtDqdDRY

### DIFF
--- a/default-modules.yaml
+++ b/default-modules.yaml
@@ -44,13 +44,13 @@ entries:
       ref: v0.5.0
 
   - name: spec-artifacts-process
-    version: "0.24.0+d605caa"
+    version: "0.24.0+375fc2a"
     defaultEnabled: true
     source:
       type: git-subdir
       url: agent-ix/spec-artifacts-process
       path: spec_artifacts_process
-      ref: d605caad70a8c09884ef22525d532e01330f763f
+      ref: 375fc2a9c49c37cce0f43384878b8ede16b162a3
 
   - name: spec-objects-business
     version: "0.6.0"
--- a/skills/spec-matrix/assets/test-matrix-example.md
+++ b/skills/spec-matrix/assets/test-matrix-example.md
@@ -14,14 +14,14 @@ This is an example test matrix demonstrating comprehensive test coverage for the
 ## Requirements Traceability
 
 ### User Story Coverage
-| User Story | Acceptance Criteria | Test Cases | Coverage Status |
+| User Story | Acceptance Criteria | Test Cases | Status |
 |------------|---------------------|------------|-----------------|
 | US-001 | US-001-AC-1 | TC-001, TC-002, TC-003 | ✅ Complete |
 | US-001 | US-001-AC-2 | TC-004, TC-005 | ✅ Complete |
 | US-001 | US-001-AC-3 | TC-006, TC-007, TC-008 | ✅ Complete |
 
 ### Functional Requirement Coverage
-| Functional Req | Acceptance Criteria | Test Cases | Coverage Status |
+| Functional Req | Acceptance Criteria | Test Cases | Status |
 |----------------|---------------------|------------|-----------------|
 | FR-001 | FR-001-AC-1 | TC-001 | ✅ Complete |
 | FR-001 | FR-001-AC-2 | TC-002 | ✅ Complete |
--- a/skills/spec-matrix/assets/test-matrix-template.md
+++ b/skills/spec-matrix/assets/test-matrix-template.md
@@ -14,18 +14,18 @@ This matrix ensures comprehensive test coverage by transforming requirements int
 ## Requirements Traceability
 
 ### Stakeholder Requirement Coverage
-| Stakeholder Req | Trace to US/FR | Test/Validation | Coverage Status |
+| Stakeholder Req | Trace to US/FR | Test/Validation | Status |
 |-----------------|----------------|-----------------|-----------------|
 | StR-XXX | US-XXX, FR-XXX | Review | ✅ Complete |
 
 ### User Story Coverage
-| User Story | Acceptance Criteria | Test Cases | Coverage Status |
+| User Story | Acceptance Criteria | Test Cases | Status |
 |------------|---------------------|------------|-----------------|
 | US-XXX | US-XXX-AC-1 | TC-001, TC-002 | ✅ Complete |
 | US-XXX | US-XXX-AC-2 | TC-003 | ✅ Complete |
 
 ### Functional Requirement Coverage
-| Functional Req | Acceptance Criteria | Test Cases | Coverage Status |
+| Functional Req | Acceptance Criteria | Test Cases | Status |
 |----------------|---------------------|------------|-----------------|
 | FR-XXX | FR-XXX-AC-1 | TC-004, TC-005 | ✅ Complete |
 
--- a/spec/evals.md
+++ b/spec/evals.md
@@ -83,7 +83,7 @@ carry a story out end to end, not whether a unit behaves. The archetype's column
 is named `Functional Req`; the ids below are the US ids each scenario drives.
 Per-FR coverage lives in TM-001 (`spec/matrix.md`).
 
-| Functional Req | Acceptance Criteria | Test Cases | Coverage Status |
+| Functional Req | Acceptance Criteria | Test Cases | Status |
 | --- | --- | --- | --- |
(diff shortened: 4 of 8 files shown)


kreneskyp:
## A caveat this fix carries is about to lose its only home

`deff9c3` recorded a caveat on this fix as a `LEDGER` entry in `src/measurement/tool-defects.ts`:

> "`traceability.status.column` named `Status` while the coverage tables asserted `Coverage Status`, so the complete-but-unbacked check never ran. Fixed in the contract by spec-artifacts-process#87 (375fc2a), which collapsed the two names. **The entry stays because the fix is per-repository: the check still does not run on any `tests.md` that has not renamed its header, and a 'no status lies' claim from one of those is unmeasured, not clean.**"

That file is deleted by [#388](https://github.com/agent-ix/quoin/issues/388), which disposes of the corpus-measurement subsystem. The deletion is correct and unrelated to this fix — but it takes the caveat with it, so it is recorded here, attached to the fix it qualifies.

**The caveat is still true.** Renaming in quoin is not renaming everywhere.

### Measured

| | |
|---|---:|
| repositories under `~/dev` containing a `type: TestMatrix` document | **219** |
| …with at least one table still headed `Coverage Status` | **138** |
| documents still carrying it | **211** of 894 |

Population counted: every `.md` declaring `type: TestMatrix` under `~/dev`, excluding the `worktrees/` scratch directory and non-repository paths. Most-affected repositories:

```
spec-artifacts-process       32 of 32
filament-ide-rs              13 of 13
quire-spec-language           8 of 8
quire-specification           6 of 6
assurance-local-evidence      6 of 6
engineering-assurance         3 of 151
quire-rs                      2 of 153
```

**`spec-artifacts-process` itself is 32 of 32** — the module that ships the collapse still has the old header in every one of its own matrices.

### What this means for those 138

Status classification is **silently skipped** on every unrenamed table, so a `status_lies: []` result from one of them means *the check never ran*, not that the rows are honest. That is the same failure this fix repaired inside quoin.

I have **not** split "stale because unrebased" from "genuinely unrenamed", and the split is not determinable from the repositories: a tree validates fine when its installed module set is equally old, and module installs live in `~/.ix` per machine rather than in the repository. Two machines can disagree about whether the same commit passes. So the 138 is a total, stated as a total.

### The cure, because it is the part that gets got wrong

**Rebase and `quoin module ensure-defaults`. Never rename toward `Coverage Status`.**

I nearly reverted this fix on exactly that reasoning — a note describing the two columns as distinct, held against a stale module set that agreed with a stale tree. Both copies I checked were pulled copies of the same age, so they agreed with each other and disagreed with the remote, and "I checked both directions" was one direction wearing two coats. The repository was ahead the whole time.

A sweep of the 138 is a separate piece of work and is not scoped here.

Issue #383, opened 2026-09-12
[Quoin#383] [Stage 5] quoin-auditor: auditor and advisor as one crate


kreneskyp:
Port `src/auditor/` and `src/advisor/` (1,806 lines after `combinatorial.ts` is extracted) to **one** Rust crate `rust/quoin-auditor`.

## The plan said two crates. The code says one.

#373 lists auditor and advisor as separate Stage 5 items. Measured against the code, the coupling is **thin but inseparable**:

- `src/advisor/advise.ts:18` → `parseSpace` from auditor — **value**, one call site. Removed entirely by the `combinatorial` extraction (see its ticket).
- `src/auditor/audit.ts:35` → `MethodCatalog` from advisor — **type only**, used at `audit.ts:101, 580, 598, 623`.

So after `combinatorial` lifts out, a two-crate split is *legal*. It is still not worth it: `audit()` takes `catalog?: MethodCatalog` as an input field (`audit.ts:101`), and **every** consumer imports both in the same file — `src/commands/evidence/audit.ts:6-7`, `src/commands/evidence/baseline.ts:6-7`, `src/commands/assurance.ts:7-8`. Nothing ships advisor without auditor. A second crate boundary would buy a seam no consumer uses.

**Recommend one crate.** If the owner prefers two, it is legal post-extraction — but the reason should be recorded, because it is not a constraint.

## Blocked by the evidence/change-assurance cycle

`src/auditor/audit.ts:21` imports `MUTATION_SCORE_METRIC, assessIndependence, scanIsVacuous` from `../evidence/index.js` at **value** level. `evidence` and `change-assurance` are a genuine value-level cycle (#376). **This crate cannot compile until that is resolved** — this is a hard ordering constraint, not a preference. See the comment on #376 for the way the fix can be done such that a third cycle survives it.

## Seams
- **No subprocess, no network.** `src/auditor/audit.ts:14` states the invariant: *"The auditor runs nothing. It reads the store and reports (ADR-0011 invariant 1)."* Preserve that in Rust — the auditor type should make execution unrepresentable, not merely unused.
- One filesystem seam: `src/advisor/methods.ts:81` reads `manifest.yaml`, transitively through `src/catalog.ts:58 defaultModuleRoots` (reads `QUOIN_MODULE_PATHS`, `readdirSync`) and `catalog.ts:152 locateModuleRoot`. One trait: `ModuleCatalogSource`.

## Stays quoin-specific — do not generalize
[...]

Issue #373, opened 2026-09-12
[Quoin#373] EPIC: Rust burn-down of Quoin first-party engine and qualification logic (staged port behind a stable boundary)


kreneskyp:
## Directive and authority

Owner decision 2026-09-12, recorded as **Amendment 1** to the campaign
[implementation language policy](https://github.com/agent-ix/quire-research/blob/main/implementation-language-policy.md)
(quire-research@8d86f64). Amendment 1 withdraws the 2026-09-07 "Quoin is retained,
no wholesale port" clause and supersedes the port prohibition in
[LR06 #62](https://github.com/agent-ix/quire-research/issues/62), which is re-scoped
to the containment boundary only.

`engineering-assurance` is the completed reference port (ADR-002, StR-003,
FR-014..FR-019, NFR-005; all Python lanes retired). This EPIC applies the same
pattern to Quoin.

## Relationship to other epics

- **quire-research [#56](https://github.com/agent-ix/quire-research/issues/56)** is the
  **containment** program: stop the spread, finish the inventory and matrix, ship
  boundary enforcement. It stays open and is not this EPIC's parent.
  This EPIC is the **burn-down**: retire the debt containment identified.
  Neither program may report the other's work as its own.
- **quoin #286** owns *semantic module contracts and staged Quire/Quoin adoption*.
  This EPIC owns *implementation language*. They collide on catalog and module code
  if that split is not held — it is stated here deliberately.

## Scope

First-party engine, production, planning, validation, canonicalization/digest,
oracle, evidence and qualification logic in `quoin`, plus corpus tooling
consolidation. Current baseline:

| Area | Lines |
|---|---|
| `src/` TypeScript | 27,723 |
| `tests/` TypeScript | 33,707 |
| `scripts/` mjs/js | 13,485 |
| `evals/` | 5,226 |
| `corpus/` Python | 4,996 |
| `templates/` | 1,767 |
| `smoke/` | 509 |

Domain sizes in `src/`: measurement 7,851 · evidence 3,958 · commands 3,891 ·
change-assurance 2,023 · assurance 1,723 · graph-analysis 1,373 · quire 1,342 ·
semantic 1,323 · auditor 1,139 · advisor 862 · completeness 685 · validators 164.

**Out of scope:** UI TypeScript anywhere; `filament-core-data`-published TS types;
all Filament cloud repos and services (frozen, no tickets); the oclif *shell* until
[...]

PR #397, opened 2026-09-12
Break the two quoin dependency cycles before any Rust crate exists (#376)

### LINKED ISSUE #376: [Quoin#376] [Stage 0 prereq] Break the two TypeScript dependency cycles blocking crate extraction
Parent: #373. Blocks #375 and every stage after it.

Cargo forbids crate cycles. Two mutual cycles exist in `src/` today and must be broken before any crate boundary can exist. This is a refactor ticket, not a translation ticket — no Rust is written here.

## Cycle 1 — evidence <-> change-assurance

- `src/evidence/index.ts:142` re-exports from `../change-assurance/schema-assets.js`
- `src/change-assurance/store.ts:22` imports `storeRoot` from `../evidence/store.js`

## Approach

Extract the shared surface downward rather than cutting an edge arbitrarily. `storeRoot` and the schema assets are both candidates for a lower module that neither side owns — this is the shape the eventual `quoin-store` crate takes, so the TS refactor should anticipate it rather than fight it later.

## Acceptance

- no import cycle between `evidence` and `change-assurance`, or between `auditor` and `advisor`; demonstrated by a check that fails if one is reintroduced
- full existing test suite green, unchanged
- no behavior change, no public API change
- the resulting module split maps onto the planned crate topology in #373

kreneskyp:
Closes #376. Part of epic #373 (Rust burn-down).

Cargo forbids crate cycles, so the cycles in `src/` are a precondition for the whole burn-down, not a translation task. Without this, #383 (`quoin-auditor`) does not compile.

## The three cycles

#376 names two. The third is the one that matters.

| # | Cycle | Edge cut |
| - | ----- | -------- |
| 1 | `evidence <-> change-assurance` | `src/change-assurance/store.ts:22 -> ../evidence/store.js` (`storeRoot`) |
| 2 | `evidence <-> change-assurance` | `src/evidence/index.ts:142 -> ../change-assurance/schema-assets.js` |
| 3 | **`auditor -> evidence -> change-assurance -> auditor`** — not named in #376 | same edge as #2 |

Cycle 3 runs through the same strongly connected component and **survives a cut that satisfies #376's wording literally**: fix `storeRoot`, leave the `evidence/index.ts` re-export in place, and a pairwise import assertion passes while `auditor -> evidence -> change-assurance -> auditor` is still alive via `src/change-assurance/types.ts:1 -> ../auditor/audit.js`. Cutting the schema-assets edge is what kills it.

`auditor <-> advisor` (cycle 4 in the SCC sense, #376's second named pair) is cut at `src/auditor/audit.ts:35 -> ../advisor/index.js` (`MethodCatalog`). The surviving `src/advisor/advise.ts:18 -> ../auditor/combinatorial.js` edge is one-directional and acyclic.

## What moved

New `src/store/` becomes the single home for what both sides of the cycles were reaching across for:

- `src/store/paths.ts` — `storeRoot`
- `src/store/canonical.ts` — `canonicalJson` / `writeCanonical`, ten callers repointed
- `src/store/integrity.ts` — JCS canonicalization, blake3, the strict I-JSON parser (moved whole from `change-assurance/`)
- `src/store/schema-assets.ts` + `src/store/schemas/` (moved from `change-assurance/`); `evidence/index.ts` re-exports from there
- `src/method-catalog.ts` (was `src/advisor/methods.ts`)

## The gate

`scripts/check-import-cycles.mjs` runs **Tarjan SCC over the whole `src/` import graph** and fails on any component larger than one module. Not a pairwise assertion — a pairwise assertion is exactly what would have let cycle 3 through. No new dependency. Wired into `make lint`.

## The bypass probe — the check was demonstrated to FAIL

A check that has never been shown to fail is a check passing over an unproven population. Each removed edge was reintroduced by hand on this tree and the gate re-run:

| Reintroduced edge | Result |
| ----------------- | ------ |
| `evidence/index.ts -> ../change-assurance/schema-assets.js` **alone** — *the deceptive partial cut that satisfies #376 literally* | **exit 1**, `component: auditor, change-assurance, evidence` |
| `auditor/audit.ts -> ../advisor/index.js` | **exit 1**, `component: advisor, auditor` |
| `change-assurance/store.ts -> ../evidence/store.js` **alone** | **exit 0** — see correction below |
| both `evidence` edges together | **exit 1**, `component: auditor, change-assurance, evidence` with all four edges listed |

**Correction to an earlier claim:** the `storeRoot` edge reintroduced *on its own* does **not** produce a cycle, and the gate correctly reports none. With the schema-assets edge already cut, `change-assurance -> evidence` is one-directional. The pair cycle needs both edges; the three-module cycle needs only the schema-assets one. So "fails on each of the three edges individually" was too strong — it fails on the two edges that individually close a cycle, and correctly stays green on the one that does not. Reported rather than restated on trust.

**Known gate limitation, filed rather than papered over:** the specifier regex matches `from "..."` and `import("...")` but not a bare side-effect `import "x";`. Such an import is a real module dependency and would be a real crate cycle, and is currently invisible to the check. This tree contains none; it is a gap to close, not a defect this PR introduces.

## Verification actually run

`make lint` — passes, including the new gate:

```
corepack pnpm run lint
$ pnpm run typecheck && node scripts/check-import-cycles.mjs && eslint "src/**/*.{ts,tsx,js,jsx}" "tests/**/*.{ts,tsx,js,jsx}" && pnpm run format:check
$ tsc --noEmit -p tsconfig.json
check-import-cycles: 27 src modules, no cyclic component
$ prettier --check .
Checking formatting...
All matched files use Prettier code style!
```

Test leg run directly (`pnpm exec vitest run`), since `make test` cannot run here — see below:

```
Test Files  106 passed (106)
      Tests  1188 passed (1188)
```

Honest note on the first execution: it reported `Test Files 2 failed | 104 passed (106)` / `Tests 3 failed | 1185 passed (1188)`, all three being `Error: Test timed out in 5000ms` in cookiecutter-render tests (`tests/semantic-module-template.test.ts`, e.g. TC-1466). An immediate rerun with no change to the tree was fully green, as quoted above. These are load-sensitive timeouts, not assertions, and are unrelated to an import-graph refactor.

## `make test` cannot run on any feature branch — pre-existing, not caused by this change

Both verified on this tree, not asserted:

**1. The verification stack is pinned to a locked revision.** `scripts/verification-stack.mjs` asserts each checkout's HEAD equals the revision in `quality/verification-stack-lock.json`:

```js
throw new Error(`${name} HEAD ${head} does not equal locked ${locked.revision}`);
```

The lock pins quoin at `8a94487074251cdba84b8f8756a15a9f54a5c23f`; this branch is at `6c1990a`. It therefore fails by design on any unmerged work. In this worktree it fails one step earlier still, on a sibling checkout the harness expects:

```
Error: quire checkout is missing at /home/peter/dev/quoin/.worktrees/quire-rs
Review the source change and prepare a candidate with make verification-relock; see quality/verification-stack.md. Relocking does not waive source checks or promote evidence.
```

The lock file was **not** edited and no gate was relaxed.

**2. `make test-with-quire` fails at its `validate` leg on a defect already on `main`:**

```
/home/peter/dev/quoin/.worktrees/376-cycles/spec/assurance/AP-201-finding-quality.md: line 43: duplicate heading 'Impact Scenarios' at level 2 (heading text must be unique per level) [duplicate-heading]
1 document(s) failed structural validation
make: *** [Makefile:226: validate] Error 1
```

Confirmed pre-existing: `git show main:spec/assurance/AP-201-finding-quality.md` has `## Impact Scenarios` at both line 34 and line 43. This diff touches **no** `spec/` file (see the diffstat — only `src/`, `scripts/`, `tests/`, `package.json`). The fix is already open as **#415**, whose only changed file is `spec/assurance/AP-201-finding-quality.md`. Not worked around here.

## Follow-up filed

`src/change-assurance/store.ts` and `src/measurement/atomic-file.ts` are not the same operation and were deliberately **not** consolidated in this PR — see #422.


kreneskyp:
Reviewed. **MERGEABLE WITH FIXES.** Both cycles are gone. The guard that is this PR's durable deliverable is blind to 12 of the 27 units it claims to cover, and that has to be fixed before merge — the cycles are a one-time cleanup, the guard is the thing that keeps them gone.

**Independently recomputed the SCCs** with my own resolver (`.js`→`.ts`, `index.ts`, dynamic `import()`), not the PR's script, on `fix/376-break-dependency-cycles`:

```
File-level graph (172 .ts files, 1 non-TS specifier): 0 SCCs size>1, 0 self-loops
Unit-level graph (27 top-level units):                0 SCCs size>1
```

Tarjan over the whole graph, not pairwise assertions — a pairwise check misses a cycle that runs through a third module, which is how both of these survived as long as they did. `pnpm run lint` passes clean.

### Must fix before merge

**`scripts/check-import-cycles.mjs:56` — `unit()` strips only `/\.ts$/`, but specifiers carry `.js`.** So `../method-catalog.js` yields the unit `"method-catalog.js"` while the file itself yields `"method-catalog"`, and the two never join. Twelve phantom nodes result: `method-catalog, catalog, version, base, modules, config-schema, flow-command, plugins, write, flows, cli, org` — every top-level leaf file is invisible as an import target.

This was confirmed empirically, not inferred: appending `import { auditSpec } from "./auditor/index.js"` to `src/method-catalog.ts` — which `src/auditor/audit.ts:35` already imports — creates a real `method-catalog ↔ auditor` cycle, and the check still prints "no cyclic component" and exits 0.

The sharp part: **this PR moved `method-catalog.ts` into that blind spot.** The file whose relocation breaks the second cycle now sits in the twelve nodes the guard cannot see.

Fix is one line — `parts[0].replace(/\.(ts|js)$/, "")`.

**Then test the guard.** `scripts/` is outside vitest's `include: ["src/**"]`, and the PR's evidence is "each cycle was reintroduced by hand", which is exactly the class of evidence that does not survive the next refactor. A fixture-graph test — a small synthetic graph with a known cycle, asserted to fail — would have caught the finding above. Without it, this check passes over a population it silently truncated, which is the empty-population failure under a different name.

### Verified, no action

**The change is relocation, not inversion, and the bodies are content-preserving.** `git diff` of the moved code is empty: `src/store/integrity.ts` and `src/store/schema-assets.ts` are byte-identical to their `change-assurance/` originals, and `src/method-catalog.ts` differs from `src/advisor/methods.ts` only in two doc sentences and one import path. `storeRoot`/`canonicalJson`/`writeCanonical` were lifted verbatim out of `src/evidence/store.ts` into `src/store/paths.ts` and `src/store/canonical.ts`. Ownership moved; logic did not — which is what a cycle break should look like, and it means the existing suite is a valid oracle for it.

**Public surface unchanged.** `src/index.ts` is byte-identical to `4d27dcf`; `evidence/index.ts:46-47` and `change-assurance/index.ts` re-export the same names from the new paths. The one widening is `writeCanonical`, previously file-private in `evidence/store.ts`, now `export`ed from `src/store/canonical.ts` — but not re-exported from any barrel, so it is internal-visible only. Acceptable.

**The four failing tests are environmental**, not caused by this PR: `tests/semantic-module-template.test.ts` and `tests/verification-relock.test.ts` fail on a stale installed `quire` CLI and an `ETIMEDOUT`, in files this branch does not touch. Installed quire modules going stale is a known repeat offender — check the pin before blaming the branch.

Fix the two findings above and this merges. It is the prerequisite for every crate boundary in the workspace, so it should not sit.

kreneskyp:

kreneskyp:
Fixes pushed as `6c1990a`. Verified against `git ls-remote`.

Both findings are closed, and one correction to my own review.

**The guard fix.** `unit()` now strips both extensions. The bug is reproduced and the reproduction is recorded: planting `import { auditSpec } from "./auditor/index.js";` in `src/method-catalog.ts` — a real cycle, since `src/auditor/audit.ts:35` already imports back — makes the guard report `27 src modules, no cyclic component` and exit 0 before the fix, and name both edges with file and line and exit 1 after it. That is the proof the finding needed: not that the fix is present, but that the guard was blind and is no longer.

**The test.** Six cases, fixture graphs built under `mkdtemp` with teardown, cycles known by construction. None runs over the real `src/` tree, which was the point — a test asserting only that the real tree is clean passes over a population the guard may have silently truncated, which is the exact bug here. Three of the six close only when a `.js` specifier resolves onto its `.ts` file, and reverting the one-line fix fails exactly those four cases while the two acyclic controls stay green. That last part matters: it shows the fix adds the missing edges without inventing any.

**Correction to my review.** I wrote that `scripts/` sits outside vitest's `include: ["src/**"]` and that covering it would need a config change. That is `test.coverage.include` in `vite.config.ts`, not the collection include. Collection uses the default, which already picks up `tests/**` — `tests/span-breadth.test.ts` imports `scripts/verify-span-breadth.mjs` exactly this way today. No config change was needed and none was made.

The script was restructured to take its source root as a parameter rather than close over a module-level constant, with the CLI behind a main guard — following `scripts/verify-span-breadth.mjs`, which is the repo's existing idiom for a script that is also imported. CLI output is unchanged.

`pnpm run lint` clean. Rebase onto `df70470` still applies before merge.

Issue #375, opened 2026-09-12
[Stage 0] Rust workspace, boundary contract and CI for the burn-down


kreneskyp:
Parent: #373. Stage 0 of the staged port.

Contains TypeScript growth before anything moves. No logic is ported in this stage.

## Deliverables

1. Cargo workspace at `quoin/rust/`, beside the retained `src/`. Not a new repo — FR-018's one-candidate-revision rule makes a second repo a pin-and-sync drift problem.
2. `quoin-core` binary skeleton: subprocess JSON-IPC dispatch, no domain logic. Reads JSON on stdin, writes JSON on stdout, `quoin-core <domain>.<op>`. stdout = payload, stderr = diagnostics, exit taxonomy 0/1/2/3/4 with "non-zero but valid payload" distinguishable.
3. `src/core/exec.ts` — the TS caller, a copy-edit of the hardened `src/quire/exec.ts`, carrying `QUOIN_CORE` / `QUOIN_EXPECTED_CORE_SHA256`. All four incidents `quire/exec.ts` encodes recur with any subprocess: executable realpath resolution, bytes pinning, 64 MiB `maxBuffer` (#164), and the three-way termination taxonomy.
4. `quoin-difftest` dev-only differential harness: same JSON request to the TS entry point and to `quoin-core`, comparing canonical-JSON stdout, exit code and normalized diagnostic shape.
5. Repo Rust idiom doc + gates: `.claude/skills/rust-style/SKILL.md`, `clippy.toml`, `deny.toml`, `rustfmt.toml`, root `[workspace.lints]` with `unsafe_code = "forbid"` and clippy `-D warnings`. ~82k lines of Rust are headed into this workspace; the idioms are Stage 0 work, not a later tidy-up. Model on `filament-ide-rs`'s.
6. CI runs both trees at one candidate revision.
7. `.language-allowances.yaml` — the exception manifest (owned jointly with the inventory work).

## Acceptance

- `cargo build`, `cargo clippy -D warnings`, `cargo deny check` green in CI
- `quoin-core` answers one trivial round-trip op end to end from `src/core/exec.ts`
- difftest runs in CI and reports on that one op
- rust-style doc exists and is cited by the review checklist
- zero TypeScript deleted in this stage

kreneskyp:
## Merge hazard from #379 — read before rebasing either branch

[...]

Issue #164, opened 2026-08-21
No maxBuffer on the quire subprocess: six commands die above 1 MiB, and the error blames unrelated stderr


kreneskyp:
`src/quire/exec.ts` shells out to `quire` with no `maxBuffer`, so Node's 1 MiB default applies. Any repository whose `quire coverage --json` payload exceeds that limit cannot run **six** commands: `assurance`, `advise`, and `evidence audit` / `record` / `affirm` / `baseline`.

The second half of this issue is that the resulting error names the wrong cause, which is what makes it expensive rather than merely annoying.

## Reproduction

`agent-ix/filament-ide-rs` at `754ae00` — 268 spec files, 1,107 obligations.

```
$ quire coverage --scope . --json | wc -c
1090714                       # exit 0, valid JSON

$ quoin assurance
Error: quire coverage --scope . --json exited abnormally:
DuplicateArchetype: 'ADR' contributed by modules ["spec-artifacts-process", "spec-artifacts-process"]; first-wins
DuplicateArchetype: 'Plan' ...
$ echo $?
1
$ quoin assurance | wc -c
0
```

1,090,714 bytes against a 1,048,576-byte limit — **4% over**.

## Cause, verified directly

Three call sites, none setting `maxBuffer`:

- `src/quire/exec.ts:29-32` — `runQuire`
- `src/quire/exec.ts:62-65` — `runQuireAllowFailure`
- `src/quire/exec.ts:82` — `quireVersion`

Measured against the real payload:

```
default maxBuffer (1 MiB): THREW code=ENOBUFS status=null signal=SIGTERM stdoutBytes=1048576
maxBuffer 64 MiB:          OK, stdout bytes=1090714
```

`advise` is doubly exposed — `src/commands/advise.ts:179` also runs `quire properties --json`, a larger payload again.

Importers of the helper: `src/commands/advise.ts` (`:74`, `:184`), `src/commands/assurance.ts` (`:65`), `src/commands/evidence/audit.ts` (`:85`), `.../affirm.ts` (`:68`), `.../record.ts` (`:115`), `.../baseline.ts` (`:61`).

## The error message reports the wrong thing

`src/quire/exec.ts:33-40`:

```ts
} catch (cause) {
  const err = cause as { stderr?: string | Buffer; status?: number };
  const stderr = String(err.stderr ?? "").trim();
  throw new Error(
    `quire ${args.join(" ")} exited ${err.status ?? "abnormally"}` +
[...]

PR #398, opened 2026-09-12
quoin-quire: collapse the quire adapter behind a Cargo edge (#379)

### LINKED ISSUE #379: [Quoin#379] [Stage 1] Collapse the quire adapter into quoin-quire
Replace `src/quire/` with a Rust crate `rust/quoin-quire` that depends on `quire-rs` directly.

Highest leverage stage: it shrinks every downstream stage, and `src/quire/contract.ts`'s vendored-schema machinery dissolves into a Cargo edge rather than being ported.

`src/quire/exec.ts` is retained until the very end — it is the template the new boundary was copied from, and it is deleted in Stage 8.

## Acceptance
- `quoin-quire` covers every capability `src/quire/` exposes
- the vendored schema + hash-assert machinery in `contract.ts` is replaced by a Cargo dependency, not reimplemented
- the 64 MiB payload from #164 round-trips (it doubles as the volume test)

---

kreneskyp:
Closes #379. Stage 1 of the burn-down.

`rust/crates/quoin-quire/` replaces the TypeScript quire adapter surface with a direct dependency on `quire-rs`, which dissolves `src/quire/contract.ts`'s vendored-schema machinery into a pin:

```
quire-rs = { version = "=0.46.0", git = "…", rev = "85dfe9d5a937c52af6456f2e6aa3a6bc4c82db9f" }
```

## Shape

`error.rs` (21 variants, codes `QQ-1001..QQ-1081`), `ids.rs`, `engine.rs`, `payload.rs`, `modules.rs`, five capability modules, `build.rs`.

Capability witnesses follow the quire-cli #68 pattern — `const _` items naming engine symbols, so the capability list cannot claim a surface the linked engine does not have.

Volume is covered by a real payload: `tests/fixtures/coverage-volume/filament-ide-rs-coverage.json`, 1,632,821 bytes — the class of payload that produced the `maxBuffer` ENOBUFS in #164.

## Two defects the tests found, both fixed and pinned

1. `Context::insert` refused a duplicate dimension *after* `BTreeMap::insert` had already displaced the value — a refusal that half-applied.
2. An assurance premise test passed for the wrong reason.

## BREAKING — cutover consequence: every stored artifact is refused

**Read this before adopting the crate.** It is deliberate, it is pinned by `tc_379_103`, and it is not changed by this PR.

`engine.rs` sets `MINIMUM_STORED_ENGINE = ENGINE_VERSION`, and for an artifact whose provenance is a 40-hex **revision** it requires equality with `ENGINE_REVISION`. This build links `85dfe9d5a937c52af6456f2e6aa3a6bc4c82db9f`. Every coverage, clause-binding and assurance artifact currently in an evidence store was written by `quire-cli` at `a874fb641cb70da83c8c8b23f9fea0a44255b88a`.

**Therefore every artifact already in the evidence store is refused by `engine::check_premise` with `EngineRevisionMismatch` the moment a consumer moves to this crate.** Not a subset — all of them.

Why it is right:

- `contract.ts` had to accept a **range** of producers, because the producer was whatever `quire` binary was on `PATH`. `minimumCli: "0.21.0"` was that range, and its own comment records it as a contract floor rather than a capability floor that had already drifted once.
- This build has exactly **one** engine. An artifact from another one was measured by a different instrument, so it is **re-measured, not reinterpreted**. A revision is not ordered — a side branch is neither newer nor older — so "accept anything at least as new" is not available even in principle.

What it costs: adopting this crate requires re-running the measurements behind every stored artifact. There is no migration path and none is planned.

What it does *not* do: it does not refuse to **read** those payloads. `payload::from_slice` parses them fine (`tc_379_100` reads the 1.6 MB fixture end to end). `check_premise` is a separate, explicit call; only a consumer that asks about provenance meets the refusal.

The same note, with the reasoning in full, is in `rust/crates/quoin-quire/tests/fixtures/coverage-volume/PROVENANCE.md`.

## Workspace membership

The crate lives at `rust/crates/quoin-quire/`, claimed by the root's `members = ["crates/*"]` glob (#375, `df70470`). Its standalone `[workspace]` table and private `Cargo.lock` are gone, and it inherits version/edition/rust-version/license/repository/authors/publish, `[lints] workspace = true` and serde/serde_json/thiserror from the root. `quire-rs` stays declared locally — it has one consumer and `build.rs` reads that exact manifest line — with the `rev` and `version = "="` pin unchanged.

No workflow file is added: main's existing `rust`, `rust-deny` and `difftest` jobs run `make rust-lint` / `rust-build` / `rust-test` over the workspace and now cover this crate.

`deny.toml` gains three scoped entries, each with its reason written at the site: `allow-git` for the quire-rs repository (which the file already anticipated), a per-crate `[[licenses.exceptions]]` admitting AGPL-3.0-or-later for quire-rs alone, and four EXACT-version `[bans] skip` entries for duplicates that arrive entirely through quire-rs's `jsonschema 0.18.3`.

## Review follow-ups from #398

- **Host-dependent oracle.** `tc_379_050` used `ModuleSelection::Ambient`, so which of three errors it met depended on the modules installed on the machine. Now a `Closed` set naming the checked-in fixture module, asserting one outcome: `DocumentRootMissing`.
- **Scratch trees leaked.** `testing::scratch` cleaned on entry and never on exit. Now a `Scratch` RAII guard whose `Drop` removes the tree, citing quoin#184 at the site, with `tc_379_025` asserting the teardown.
- **Unanchored pin parser.** `build.rs`'s `field()` split on the bare key, correct only because no current URL contains `rev` or `version`. Now anchored on a TOML token boundary followed by `=`. Since a build script is never compiled as a test target, the parser moved to `build_support/manifest_pin.rs`, `include!`d by both `build.rs` and `tests/manifest_pin.rs`; `tc_379_027` feeds it a line whose git URL carries both words ahead of the real keys.
- **Cutover consequence undocumented.** The section above, and PROVENANCE.md.

Review finding 2 — `FR-096`–`FR-100`, `NFR-024`, `NFR-025` do not exist yet — is **not** addressed here. Authoring them is #410; the Trace tags are left exactly as they are.


kreneskyp:
Reviewed. **MERGEABLE WITH FIXES.** Two must land before merge; four are follow-ups I am not holding the branch for.

The core question for this PR was whether `src/quire/exec.ts`'s four hard-won incidents survive the collapse. They do, and the evidence is specific:

| incident | disposition |
|---|---|
| executable realpath resolution | correctly **retired** — `src/lib.rs:22` "no path to resolve"; there is no `PATH`/`realpath` surface left, and `src/quire/exec.ts` is retained untouched per FR-018 |
| `QUOIN_EXPECTED_QUIRE_SHA256` bytes pinning | **retired with a stated replacement** — `rev = "85dfe9d5…"` plus `version = "="` in `Cargo.toml:28`, enforced by `build.rs:44-62` (rejects branch/tag refs, requires 40 hex) and pinned by `tc_379_020` against the manifest text |
| 64 MiB `maxBuffer` (#164) | **preserved and re-scoped** — `PayloadLimit::DEFAULT` at `src/payload.rs:80`; `read_bounded` reads `limit+1` so overflow is distinguishable from truncation; `tests/volume.rs` exercises a real 1,632,821-byte payload with checked-in provenance and no `node`/`quire` at test time |
| three-way termination taxonomy | **preserved** — `DocumentOutcome::{Pass,Fail,CouldNotRun}` at `src/validate.rs:75` with the `failure` discriminator on the wire; the `runQuireAllowFailure` case is `properties::Outcome{report, unresolved}` |

Retiring an incident is allowed; retiring it silently is not. Both retirements here name what replaced them, which is the standard.

`contract.ts` genuinely dissolves rather than relocating: no `*.schema.json`, no hash table, no `minimumCli` anywhere under `rust/quoin-quire/`. Payloads deserialize into `quire_rs` types through `pub mod model` re-exports. That is the Cargo edge doing the work the vendored-schema machinery used to do, which is the whole point of putting this stage first.

### Must fix before merge

**1. Nothing builds this crate.** `.github/workflows/build-test.yml` runs cargo only for `.ci/quire-cli`, the `Makefile` has no rust target, and `Cargo.toml:36` makes `quoin-quire` its own workspace, so even #375's workspace would not pick it up. "38 tests pass" is a local assertion. This PR's own body warns #375 about checks that pass over an empty population — and then ships a crate whose test suite no repository lane runs. Add the CI job or `make` target. `workflow_dispatch:` only, per standing policy.

**2. Every trace tag names a requirement that does not exist.** `FR-096`–`FR-100`, `NFR-024`, `NFR-025` are all absent — `spec/functional/` ends at FR-095 and `spec/non-functional/` at NFR-023 — and `spec/matrix.md` carries no `TC-379-*` row. Under quoin's own rollup the crate is 100% untraced. The plan allocated those IDs; nobody authored the documents. Either author them or retag to what exists. A tag pointing at nothing is worse than no tag: it reads as covered in every count that consumes the matrix.

### Follow-ups (file, do not hold)

3. `engine.rs` sets `MINIMUM_STORED_ENGINE = ENGINE_VERSION` and requires revision equality, so every artifact already in the evidence store at engine `a874fb64…` is refused — `tc_379_103` pins exactly that. It is deliberate and documented in code, but it is a cutover consequence and belongs in the PR body and the stage's cutover ticket, not only in a test name.

4. `src/coverage.rs` `tc_379_050` asserts over two alternative errors using `ModuleSelection::Ambient`. The oracle then depends on which modules the host machine happens to have installed. Use a `Closed` fixture set — the ambient variant is the reason installed-module staleness keeps producing false results elsewhere in this repo.

5. `src/testing.rs:14` `scratch()` cleans on entry and never on exit, leaving trees behind. That is quoin#184's pattern; reference it rather than re-deriving it.

6. `build.rs:70` `field()` extracts by naive `split(key)`. It is correct today only because neither `"rev"` nor `"version"` appears earlier in the dependency line — a URL change silently mis-parses, and this function is the thing enforcing the revision pin. Anchor on `key =`.

Findings 1 and 2 are the same defect in two places: the gate exists in the artifact but in no lane that runs it. Fix both and this merges.

kreneskyp:

kreneskyp:
Fixes pushed as `6597360`. Verified against `git ls-remote`. All four review findings closed, the crate is now a workspace member of the merged root, and the gate is green: 72 tests pass, clippy clean under `-D warnings`, `deny check` reports `advisories ok, bans ok, licenses ok, sources ok`.

One thing worth recording, because the first version of it would have quietly weakened the workspace gate.

## The `deny.toml` skip list

Adding `quire-rs` as the workspace's first git dependency pulls in `jsonschema 0.18.3`, whose `url`/`getrandom`/`icu` chain carries older majors than the rest of the graph resolves. That produces four duplicate pairs against `multiple-versions = "deny"`.

Keeping the check at `deny` with narrow exact-version skips, rather than dropping it to `warn`, is the right shape and I am accepting it. A skip names one resolution; `warn` disables the check for every future duplicate as well.

But the first version skipped the wrong member of the `syn` pair, and the justification comment asserted a provenance that had not been verified. I traced all four with `cargo tree -i --target all --edges normal,build`:

| skipped crate | reachable from |
|---|---|
| `getrandom@0.2.17` | jsonschema 0.18.3 → quire-rs |
| `getrandom@0.3.4` | ahash → jsonschema 0.18.3 → quire-rs |
| `r-efi@5.3.0` | getrandom 0.3.4 → ahash → jsonschema 0.18.3 → quire-rs |
| `syn@2.0.119` | **serde_derive 1.0.228 → quoin-core and quoin-difftest, directly** |

The `syn` duplicate pair is `2.0.119` vs `3.0.5`, and only `3.0.5` is quire-rs-only — it arrives through displaydoc → icu → idna → url → jsonschema. `2.0.119` is what the workspace's own `serde = "=1.0.228"` pin pulls. Skipping it would have exempted the half of the graph we control, which is the half the check exists to police: a future direct dependency of ours introducing a second `syn 2.x` would then have passed silently.

Now corrected to `syn@3.0.5`, and the gate passes with our own `syn` checked — so the duplicate is genuinely one-sided, which is the thing that had to be true for a skip to be honest here.

**The rule the list now follows, stated in the file:** for each pair, the skipped side is the one reachable only through `quire-rs`; the other side stays checked. Every entry carries its full provenance chain, and the comment names the command the chain was verified with.

The general form — worth carrying to the other burn-down branches. A dependency-gate exemption is only as good as its justification, and a justification nobody re-derived is not evidence. The original comment also said "these three resolutions" above four entries, which is the kind of drift that tells you the prose and the list stopped being checked against each other. The next reader of a `skip` list has no way to tell a verified concession from a plausible one, so the provenance has to be in the file, not in the PR that added it.

## Scoped, not general

Two other allowances landed with the git dependency, both deliberately narrow:

- `allow-git` lists the `quire-rs` repository alone, with `unknown-git = "deny"` still refusing every other git source. It does not weaken the revision pin — the crate pins by `rev` and `build.rs` enforces the SHA.
- AGPL is admitted through a per-crate `[[licenses.exceptions]]` for `quire-rs` only, so the `allow` list stays permissive-only and AGPL cannot arrive unnoticed in a future dependency.

Trace tags untouched, as instructed — the nonexistent `FR-096`–`FR-100` / `NFR-024` / `NFR-025` ids are #410's to author, not this PR's to retag.

Remaining before merge: rebase onto `df70470` is already done; this branch is mergeable once #410 authors the requirements its tags name.

Issue #396, opened 2026-09-12
[Quoin#396] The only test proving @agent-ix/filament-plan-sync still resolves dies with ix-cli-core — it needs a successor first


kreneskyp:
## The dependency

`tests/it-005-sync-discovery.test.ts` asserts quoin's oclif plugin contract — that `@agent-ix/filament-plan-sync` is discoverable and resolves — and it does so **through `listCorePlugins`, an `@agent-ix/ix-cli-core` symbol**.

So the test that proves the published extension contract still works is itself built on the dependency the burn-down retires. Retiring ix-cli-core deletes the proof at the same moment it changes the thing being proved.

Found during #381 (config/modules port), which deliberately touched none of it.

## Why this is a precondition, not a cleanup

#373's plan already defers the oclif decision — port `filament-plan-sync`, keep an escape hatch, or drop it — to the end of Stage 8, and holds command-surface parity meanwhile via `tests/command-entries.test.ts` and `tests/cli-usage.test.ts`. Those two snapshots cover quoin's *own* commands. Neither covers a third-party plugin resolving, which is the part quoin does not control and cannot notice breaking.

FR-018's rule is that code and its tests are deleted together. Here the test's *subject* outlives the test: the plugin contract is published, and consumers keep relying on it after ix-cli-core is gone. Deleting the test with the dependency leaves a live contract with no gate.

## What this ticket wants, in order

1. Write the successor **before** the dependency leaves `package.json`. It must prove the same thing — a third-party oclif plugin is discovered and resolves — without importing an ix-cli-core symbol.
2. Decide what the successor asserts against once oclif is retired. If the answer is "nothing, because the contract is withdrawn", that is a deliberate contract withdrawal and needs the owner's decision plus a deprecation path for `@agent-ix/filament-plan-sync`, not a silent deletion.
3. Only then delete `it-005-sync-discovery.test.ts`, in the same commit as the code it covers.

## Constraint from #381

[...]

Issue #378, opened 2026-09-12
[Quoin#378] [Stage 3] Port semantic and completeness to Rust crates


kreneskyp:
Port `src/semantic/` and `src/completeness/` to `rust/quoin-semantic` and `rust/quoin-completeness`.

Both are leaves with no dependents, which is why they come before the evidence tree. **This is where the ajv vs `jsonschema` crate divergence gets discovered** — on a small blast radius instead of inside measurement.

## The known risk
`src/semantic/manifest.ts`'s `mapAjvError` and `src/quire/validate.ts` both surface ajv error shape and ordering to users. The Rust `jsonschema` crate will not reproduce it. Ecosystem pins also disagree (quire-rs ~0.18 vs quire-contract-ir 0.17) — resolve that as part of this ticket.

## Acceptance
- golden corpus: every schema x ~50 valid/invalid documents
- **exact verdict parity**; normalized (path, keyword, instance-location) diagnostic parity
- the divergence surface is documented, not silently absorbed

---
Parent: #373. Depends on the workspace from #375 for wiring, but the crate itself is written and tested standalone — do not block on it.

Standing rules: `/rust-review` is the style guide, a repo `rust-style` doc outranks it. `#![forbid(unsafe_code)]`, one `thiserror` error enum per crate boundary with stable codes, newtypes over bare `String`/`Uuid` for ids. The existing TypeScript is the oracle: capture expected fixtures from it once, never leave TS as a runtime oracle. **No TypeScript is deleted in this ticket** — cutover and deletion are separate tickets per FR-018.

kreneskyp:
Cutover ticket opened: #445. #378 delivered the library; #445 is what closing it means — `completeness.*` reachable from the `quoin-core` binary and `src/completeness/` deleted. Gated on #412.

Issue #377, opened 2026-09-12
[Quoin#377] [Stage 2] Port validators to quoin-validators


kreneskyp:
Port `src/validators/` (~164 lines, 1 command) to a standalone Rust crate `rust/quoin-validators`.

Deliberately the smallest domain in the repo. It is first so that the boundary, error taxonomy, generated types, difftest and the delete step are each proven once on a target where a failure is unambiguously the mechanism and not the domain.

## Acceptance
- `quoin-validators` builds, clippy clean at `-D warnings`
- every behaviour asserted by the existing TS tests is asserted by a Rust test
- verdict parity with the TS implementation on a golden corpus, exact
- diagnostic parity normalized on (path, keyword, instance-location); **error text is not contractual, verdicts are**

---
Parent: #373. Depends on the workspace from #375 for wiring, but the crate itself is written and tested standalone — do not block on it.

Standing rules: `/rust-review` is the style guide, a repo `rust-style` doc outranks it. `#![forbid(unsafe_code)]`, one `thiserror` error enum per crate boundary with stable codes, newtypes over bare `String`/`Uuid` for ids. The existing TypeScript is the oracle: capture expected fixtures from it once, never leave TS as a runtime oracle. **No TypeScript is deleted in this ticket** — cutover and deletion are separate tickets per FR-018.

PR #399, opened 2026-09-12
quoin-config and quoin-modules (#381)

### LINKED ISSUE #381: [Quoin#381] [Stage 7] Port config, plugins and modules
Port `src/config/`, plugin registration and `src/modules/` to Rust crates.

Runs **parallel to the entire evidence tree** — it shares no code with it. The only stage needing git and network.

## Notes
- `src/config-schema.ts` is 56 lines and the schema is `{ org?: string }` strict. zod -> schemars is near-trivial. The real coupling is `@agent-ix/ix-cli-core`'s `ConfigService` / `registerPluginSchema`.
- **Do not port `@agent-ix/ix-cli-core`.** quoin uses 9 of its symbols and none of its ~2.3k lines of auth/secrets/marketplace. Reimplement those 9 behaviours natively and list them explicitly.
- `quoin-modules` needs git — use `gix`.

---

---

## AMENDMENT 2026-09-12 — config is CONSUMED from `ix-cli-kit`, not ported

**Everything above this line is SUPERSEDED in the `config` portion only, and is retained deliberately as the record of what this ticket said before `ix-cli-kit` existed.** It is not rewritten. Plugins and modules below are unchanged and remain in scope.

### What changed

`https://github.com/agent-ix/ix-cli-kit` was created on 2026-09-12: a private, `publish = false`, AGPL-3.0-or-later shared Rust crate that is a *library of CLI plumbing* for the IX CLIs — explicitly **not** a CLI, not a framework, and not a port of `ix-cli` (which stays TypeScript and is out of scope for this program). Its v0.1 ships `exit`, `streams`, `json`, `version` and `config`.

Its `config` module already carries the configuration-resolution behaviour this ticket was written to build:

- `SearchPath` — an ordered, deduplicated **union** of search roots, for *where to look*.
- `resolve` — the **override** chain **flag > env > file > default**, for a single *value*.
- Absent and malformed configuration are kept distinct; a malformed file is reported with path, line and column.
- `config::xdg` is available but never the imposed default — the ecosystem's real default module root is `~/.ix/filament/modules`, which quoin materialises into and quire-rs reads. **The crate decides the ORDER; the consumer decides the LOCATIONS.**

That module is a 1:1 port of `quire-cli`'s `scoped_registry_roots` precedence chain (`/home/peter/dev/quire-cli/src/commands/validate.rs:485-515`), including the `IX_FILAMENT_MODULES_PATH` / legacy `IX_SCHEMA_PATH` union and the `default_module_root()` fallback. quoin must not write a second one.

### Revised scope of this ticket

1. **Config: CONSUME, do not port.** quoin takes configuration precedence and search-path resolution from `ix-cli-kit::config` via a git-rev pin (`ix-cli-kit = { git = "https://github.com/agent-ix/ix-cli-kit", rev = "<40-hex sha>" }` — rev, never tag, never vendored, never crates.io). No config precedence implementation is written in this repository. The `zod -> schemars` note above still applies to quoin's own `{ org?: string }` config *schema*, which is quoin's payload shape, not the resolution order.
2. **Plugins and modules: unchanged, still quoin's own work under this ticket.** Plugin registration and `src/modules/` (git via `gix`) stay here. `ix-cli-kit` explicitly excludes plugin / command dispatch — its README reserves exit code `5` for `command_not_found` and deliberately leaves `exit::from_code(5)` returning `None`, gated on quoin `spec/functional/FR-102-command-surface-and-oclif-retirement.md` reaching a decided disposition **and** a second Rust consumer existing. Neither condition holds today.
3. **If `ix-cli-kit` does not fit, file a gap ticket in `ix-cli-kit`. Never reimplement locally in quoin.** A local reimplementation is the exact failure this amendment exists to prevent: it restores the two-implementations-of-one-behaviour state the shared crate was created to end. The remedy for an unsuitable upstream is an upstream ticket, not a fork of the behaviour.

### Ordering dependency — do not start this ticket yet

`ix-cli-kit` **has no spec yet.** One is being authored now in a parallel task. This ticket must not start against an unspecified dependency: building quoin's config consumption against an unspecified crate would make quoin's usage the de-facto specification, which is the same defect as porting the code here.

**Blocked until:** `ix-cli-kit` has a merged spec, and quire-research#69 (repo creation and initialisation) reaches a state that gives a pinnable rev. Plugins and modules work under this ticket is likewise held, because it shares this ticket's crate and wiring.

kreneskyp:
Closes #381. Stage 7, which runs parallel to the evidence tree.

Two crates: `rust/quoin-config/` and `rust/quoin-modules/`. 73 tests, 0 failed, 0 ignored. `cargo +1.98.1 fmt --check`, `clippy --all-targets --all-features -- -D warnings` and `test` all clean.

The gix path runs end to end against a repository the test builds object by object over the local transport — no network lane and no `#[ignore]`. Bounded blocking is a watchdog thread raising gix's `should_interrupt` at a budget and joined on drop, not an async runtime.

## The shipped ix-cli-core types lie in three places

Verified against `dist/index.js` and against disk, not against the `.d.ts`:

- The config store is **YAML at `~/.config/ix/config.d/*.yaml`**, not JSON and not under `~/.ix/`. Confirmed on disk: `~/.config/ix/config.d/local.yaml`.
- quoin imports **11** symbols, not the 9 the plan stated (10 values + 1 type across 6 files in `src/`, plus `listCorePlugins` in `tests/`).
- Behaviour and successor for each of the 11 is tabulated in `rust/quoin-config/IX_CLI_CORE_RETIREMENT.md`. Five are reimplemented here; the remaining seven are CLI shell and belong to Stage 9.

## A zod/serde parity bug the goldens caught

zod's `.optional()` refuses an explicit `null`; serde's `Option<T>` accepts it, so `{ org: null }` was silently valid on the Rust side. Fixed with a `deserialize_with` that refuses present-but-null, pinned by `tc_381_102`. `rust/goldens/PROVENANCE.md` records this and four other deliberate divergences.

## Follow-ups filed, not fixed here

- #396 — `tests/it-005-sync-discovery.test.ts` is the only test proving `@agent-ix/filament-plan-sync` resolves, and it asserts that through `listCorePlugins`. It needs a successor before the dependency goes.
- #375 — delete the two per-crate `Cargo.lock`s when wiring members; decide the `cargo deny` position on `serde_yaml_ng` → `unsafe-libyaml` against these crates' `#![forbid(unsafe_code)]`.


kreneskyp:
Reviewed. **BLOCKED** on one functional break. Everything else is fixes, two nits, and one spec question I am ruling on below.

Verified clean: only `rust/` changed, no corpus or evidence bytes touched, no `package.json` change, `publish = false` on both crates, `#![forbid(unsafe_code)]` at both roots, 42 tests green under `cargo +1.98.1 test`, clippy silent at `-D warnings`. PROVENANCE's revision and `default-modules.yaml` pin claims both check out, and there is no `sha256:`/`sha256-jcs:` conversion anywhere in the diff — the digest domains stay separate, which is what #409 exists to protect.

The tautology guard passes properly here: the goldens are captured from TypeScript rather than written by the Rust tests, every oracle loop asserts a minimum population, and the reconcile tests count resolver calls instead of assuming they happened. That last one is the difference between a test and a decoration.

### BLOCKING

**`rust/quoin-modules/src/git.rs:413` — `Revision::Head` is broken.** It resolves `refs/remotes/origin/HEAD`, which the refspecs at `git.rs:360` never create. Reproduced against the crate's own fixture repo: every unpinned source fails with `QM013_GIT_REVISION_NOT_FOUND`. No test covers `Head` at all, which is why 42 green tests did not catch it — the default path for an unpinned module source is untested.

Fix the resolution and add the case. An unpinned source is the common case, not an edge.

### Must fix before merge

These are all the same species: a difference from the retained TypeScript that is real, small, and not written down. `PROVENANCE.md` lists divergences; a divergence not in that list is a defect regardless of which behaviour is better.

1. `install.rs:313` — `installedAt` is second-precision; TS `writeRegistry` emits milliseconds (`2026-01-02T03:04:05.678Z` in the golden). Both implementations write this registry file during coexistence, so this is a shared on-disk surface, not an internal detail.
2. `config/src/service.rs:276` — `set()` drops the advisory `.lock` that ix-cli-core holds around its read-merge-write. Concurrent writes to the shared `~/.config/ix/config.d/quoin.yaml` can lose a key. Not in the divergence list.
3. `config/src/service.rs:42` — writes `# Managed by quoin …` where ix-cli-core stamps `# Managed by ix …`. The two implementations will churn each other's header on alternate writes, and `tests/config_service.rs:50` pins the new string, so the test now enforces the churn.
4. `git.rs:116` — `Builder::spawn(…).ok()` discards a watchdog spawn failure, leaving the blocking fetch with no wall-clock bound at all. A failed timeout is worse than no timeout, because the code above it believes there is one.
5. `git.rs:133` — `expired()` ORs elapsed time with the interrupt flag, so any transport failure after the budget is relabelled `QM015_GIT_TIMEOUT` and the real error is lost. This is the "never hide errors by design" case exactly: the operator sees a timeout and tunes a timeout that was not the problem.
6. `config/src/service.rs:277` — `set()` errors on an unparsable existing file where the oracle swallows and writes. `tests/config_service.rs:159` asserts the opposite of what the oracle does, so the suite is currently pinning a divergence as if it were parity.
7. `config/tests/config_service.rs:202` — `deny_unknown_fields` stops at the first unknown key; zod `.strict()` reports all of them. `config doctor` now names one of three unknown keys. Undocumented, and user-visible.
8. `modules/tests/ts_oracle.rs:146` — the only byte-parity assertion uses a record with every optional field absent, so `ref`/`sha`/`semantic` key ordering and presence is pinned by nothing. Byte parity over the empty shape is the empty-population case wearing a golden's clothes. Add a record with every optional field present.

Nits, fix while you are in there: `plugin_registry.rs:81` reports a non-strict schema as `QC001_INVALID_PLUGIN_ID` with `"quoin: schema is not strict"` fabricated into the id field — that is a different error wearing an id's error code. And `git.rs:596` silently skips symlink and submodule tree entries, a real divergence from `git checkout` that belongs in PROVENANCE.

### RULING on finding 10 — the spec conflict is real, and the answer is not to change this PR

The reviewer is right that there is a contradiction. `spec/functional/FR-027-*.md:82` AC-8 requires config handlers to **delegate to ix-cli-core**, and this PR reimplements them. FR-098-AC-1 makes `quoin-difftest` the parity vehicle, and this PR invents `rust/goldens/` plus per-crate `ts_oracle.rs` instead. Neither the spec nor the matrix changed.

Two separate answers:

**On AC-8 — the spec is not violated yet, but it will be, and that is planned.** The burn-down explicitly does not port `@agent-ix/ix-cli-core`: quoin uses nine of its symbols and none of its auth, secrets or marketplace surface, so the ruling was to reimplement those nine behaviours natively. FR-027-AC-8 describes the retained TypeScript, which still delegates and is untouched by this PR. A Rust crate that reimplements them is the intended design.

What is not acceptable is that the conflict is nowhere in writing. Recording a divergence and committing does not settle it — it preserves it. So: **AC-8 is amended at the cutover ticket, not now, and the amendment is a named deliverable of that ticket.** Add a line to `PROVENANCE.md` in this PR saying the Rust crate deliberately does not delegate, that FR-027-AC-8 governs the retained TypeScript until cutover, and that cutover must amend it. That is the whole fix here — one paragraph — and without it the next reader finds a crate contradicting a criterion with no record of anyone having noticed.

**On `quoin-difftest` — per-crate golden oracles are correct here and are not a second parity mechanism, provided they stay one shape.** `quoin-difftest` compares the live TypeScript entry point against `quoin-core` over the IPC boundary. There is no IPC in this PR and no `quoin-core` dispatch for config or modules, so there is nothing for difftest to drive. A captured-once golden set is the right instrument for an unwired library crate, and #377 and #378 reached the same shape independently.

But three branches independently inventing a golden home is how a repository ends up with three. **Consolidate: one goldens root under `rust/`, one capture-script convention, one `ts_oracle.rs` shape.** I am filing that as a Stage 0 follow-up rather than holding this PR — #401 merged to `main` a few minutes ago (`df70470`), so the workspace now exists to hang it on, and whichever of the five branches rebases first should land the convention.

Fix the blocker, the eight, the two nits and the PROVENANCE paragraph, and this merges.

kreneskyp:

kreneskyp:
## Fix verification — all blocking items closed, `7c8ffc0`

Four commits on `df70470`: the port fixes, the licence/skip policy, the `serde` bump, the `thiserror` bump + `syn` skip.

| Gate | Result |
|---|---|
| `cargo +1.98.1 fmt --all --check` | exit 0 |
| `cargo +1.98.1 clippy --all-targets --all-features --locked -- -D warnings` | clean |
| `cargo +1.98.1 test --locked --workspace` | zero failures |
| `cargo +1.98.1 deny --manifest-path rust/Cargo.toml check` | **advisories ok, bans ok, licenses ok, sources ok** |

### The `deny` policy, and why it is three rulings rather than one

The deciding fact was established by checking rather than assuming: `origin/main` passes `deny` cleanly and has no `gix`. So every rejection arrives with `gix` itself. That settles the alternative before it is asked — `uluru` and the `hashbrown` duplicates are core `gix`, not transport-specific, so dropping `blocking-http-transport-reqwest-rust-tls` for `blocking-http-transport-curl` would not remove them. It would trade a pure-Rust TLS stack for a runtime `libcurl` dependency, which is worse for a binary shipped through npm platform packages. The reqwest TLS transport stays.

**1. Four permissive licences added to `allow`** — ISC, BSD-3-Clause, Zlib, CDLA-Permissive-2.0. Not a loosening: `deny.toml` states that a new dependency adds its licence in the same commit that adds the dependency, and that the list is exactly what the current graph carries.

**2. `uluru`'s MPL-2.0 is a named per-crate `[[licenses.exceptions]]`, not an `allow` entry.** MPL-2.0 is weak copyleft and is a different kind of thing from the four above; an `allow` entry would let MPL arrive unnoticed in a future dependency. The exception records that MPL §3.3 is explicitly GPL/AGPL-compatible and quoin is AGPL-3.0-or-later, and that we consume `uluru` unmodified so §3.2 is satisfied — a policy exception to a permissive-only list, not a legal exposure.

**3. Duplicates: resolved where they are ours, skipped only where they are not.** `deny.toml`'s "a duplicate version is resolved, not conceded" means *we* resolve what *we* control, and each pair was traced with `cargo tree -i --target all --edges normal,build` rather than assumed:

- **`hashbrown`** — `0.17.1` is ours (quoin-config → `serde_yaml_ng` → `indexmap`) and stays counted. `0.14.5` (dashmap → gix-tempfile) and `0.16.1` (clru → gix-pack) skipped.
- **`base64`** — both sides gix-only. Skipped the narrower `0.23.1` (reqwest → gix-transport), leaving `0.22.1` counted, so the most of the graph stays under the gate.
- **`syn`** — this one was ours, twice, and is **resolved rather than skipped**. `2.0.119` reached quoin-config/core/modules through `serde_derive` and, hiding behind it, `thiserror-impl`. Raising `serde` `=1.0.228` → `=1.0.229` and `thiserror` `=2.0.18` → `=2.0.20` moves both onto `syn 3.x`. Only then is `2.0.119` gix-only (`gix-macros`, `jiff-static`, `jni-macros`, `jni-sys-macros`, and the `yoke`/`zerofrom` → icu → idna → url → reqwest chain), and only then is it skipped. `syn@3.0.5` stays counted.

Both pin raises are one patch ahead of `quire-corpus`, whose `=` pins the comment at `rust/Cargo.toml:53` says quoin matches. The comment is amended in the same commits to state the divergence, its reason, that `quire-corpus` has no `schemars` and so is not wrong today, and — the part that matters for future readers — that reverting **either** pin removes the `syn` skip's premise and turns it into cover for a duplicate we could resolve.

### Other items

`Revision::Head` is fixed by adding gix's own third refspec `+HEAD:refs/remotes/origin/HEAD`, rather than working around the missing ref downstream: `+refs/heads/*:refs/remotes/origin/*` creates no ref named `HEAD`. Pinned by `tc_381_297`. `PROVENANCE.md` records that `ref_map.remote_refs` is filtered by the supplied refspecs, so the next person does not spend that attempt again.

Eight items and two nits from the review are closed.

### Not closed here

Trace tags on this branch name criteria that do not exist (FR-096..FR-100, NFR-024, NFR-025). That is #410 and it blocks merge across the whole batch, not this branch alone.

kreneskyp:
Blocker resolved, and the resolution is a fourth option the train did not list.

All three options considered — edit the test's expectation, narrow the EA dev-dependency, bump a pin — were correctly refused. The actual defect is neither the test nor the dependency: **it is the bridge**. Any code that re-serialises a `serde_json::Value` through a non-`serde_json` serializer is broken whenever `serde_json/arbitrary_precision` is unified on, which is one dependency edge away at all times (EA's `default = ["full"]` enables it). That is a real bug on its own merits, it lives inside this PR's own code, and fixing it is what makes the test pass honestly.

Two sites, both fixed here:

1. `quoin-modules/tests/ts_oracle.rs` — re-serialised the golden through `serde_yaml_ng::to_string`, so `from_yaml` refused a manifest that is in fact legal. Now feeds the golden's own bytes; YAML 1.2 is a superset of JSON, so this is the more faithful oracle anyway.
2. `quoin-config/src/service.rs` — `set_key`'s `KeyKind::Complex` branch used `serde_yaml_ng::to_value(parsed)`, which would write `{"$serde_json::private::Number": "7"}` into the user's config file. Now a JSON-text round-trip.

**One correction to the train's report.** It stated that `quoin config set` currently writes the marker. It does not. `grep -rn "KeyKind::Complex"` matches exactly one line — the call site itself. No shipped schema declares a Complex key, so the branch is unreachable outside a plugin schema. The defect is **latent, not live**. Worth stating precisely, because "we shipped corrupt config files" and "we have an unreachable branch that would corrupt config files" are different facts.

Regression test `tc_381_131` builds a test-local `PluginConfigSchema` declaring a Complex key, since no shipped schema does. Per the unfixed-first rule it was run against the unfixed code first, workspace-wide, and fails with the marker written to the file:

```
limits:
  deep:
  - $serde_json::private::Number: '1'
  retries:
    $serde_json::private::Number: '7'
```

`make rust-gate` green: fmt, clippy -D warnings, cargo deny, tests, the end-to-end caller, and 72 difftest cases with 0 differences.

Follow-ups filed as quoin#442 — no gate yet stops the next such bridge, the other eleven crates are unswept, and the EA `full`-feature side is tracked separately.

PR #400, opened 2026-09-12
quoin-validators: the first end-to-end cutover (#377)

### LINKED ISSUE #377: [Quoin#377] [Stage 2] Port validators to quoin-validators

kreneskyp:
Closes #377. Stage 2.

Deliberately the smallest domain in the repo — 164 LOC behind one command. The point is not the port; it is that the first end-to-end cutover proves the IPC boundary, the exit-code taxonomy, generated types, the differential harness and the delete step without a domain surprise underneath.

Nothing here deletes TypeScript. Per FR-018 the existing implementation is retained until its Rust replacement proves equivalence at one candidate revision, and the cutover and the deletion are separate commits.


kreneskyp:
Reviewed. **BLOCKED** on two concrete defects. A third issue is a scope ruling rather than a merge blocker, and is below.

Gates re-run rather than read: `cargo +1.98.1 fmt --check`, `clippy --all-targets --all-features -- -D warnings`, and `cargo test` (15 unit + 4 integration + 1 doctest) all pass with an explicit `--target-dir`.

### BLOCKING 1 — the branch breaks the TypeScript test lane

`rust/quoin-validators/tests/golden/generate-oracle.test.ts:32` imports `../src/commands/validate.js`. Relative to `tests/golden/`, `repoRoot` resolves to `rust/quoin-validators/tests`, so the import points at `rust/quoin-validators/tests/src/commands/validate.js`, which does not exist.

The file's own comment at `:13-14` says it "lives under `rust/`, which `vite.config.ts` does not collect". I checked the config on this branch. `test.exclude` is exactly:

```js
exclude: ["node_modules/**", "dist/**", "corpus/**", ".worktrees/**"]
```

No `rust/**`. Vitest's default glob collects the file, and `npx vitest run` on the branch reports `1 failed (1)`.

The irony is that `vite.config.ts` already carries two long comments explaining this exact class of bug — `corpus/**` was excluded after vitest collected and executed the qa-corpus's `*.test.ts` fixtures and failed on `trace is not defined`, and `.worktrees/**` after it collected 2220 files instead of 97. This is the third instance of a known pattern and the comment asserting immunity was never checked against the config.

**Both halves need fixing, not just the import.** Add `rust/**` to `test.exclude`, **and** rename the capture so it is not `*.test.ts`. If only the import were corrected, the file at `:117` unconditionally rewrites `expected.json` on every routine run — converting the golden into a fixture the test writes for itself. That is the tautology guard exactly: a criterion backed only by a fixture the test authored is not evidence. The header's own reasoning is right — "a golden that is cheap to refresh stops being a gate the first time the implementation changes" — so make the file structurally incapable of running by accident, rather than relying on a claim about collection that turned out to be false.

### BLOCKING 2 — an undocumented behavioural divergence

`src/gates.rs:101-105` reads every wiring file eagerly. `src/validators/gates.ts:38-40` reads lazily, inside `find`. A repository with an unreadable `Makefile` and no claiming gate returns `[]` under TypeScript and `QV-E003` under Rust.

`PROVENANCE.md` lists two divergences and not this one. Either document it with its justification, or make the Rust lazy to match. A behaviour change bundled into a language change cannot be attributed to either afterwards — which is the same reason #407 was filed against `digest_file_sha256`.

### The differential harness is sound, and this is the part that matters most

`cases.json` is hand-authored input; `expected.json` was captured from the retained TypeScript at `4d27dcf`. The reviewer independently recomputed all three source hashes in `PROVENANCE.md` against `origin/main` — they match, and `4d27dcf` is current `main`. 46 cases, 23 with findings, 23 clean, 28 findings total. `tc_377_019` compares the full finding payload, byte-exact `--json` stdout, human lines, and strict exit per case; `tc_377_020` guards the floor so a trimmed corpus fails. No Node runs at Rust test time.

That is a real oracle over a non-empty population, captured once and frozen. It is what the rest of the stages should copy.

### RULING — this crate does not demonstrate the boundary, and the stage is not complete when it merges

There is no `[[bin]]`, no `quoin-core`, no subprocess, no stdout/stderr split, no `schemars`, no generated TS types, no hash assertion. The `quoin-core <domain>.<op>` contract exists here only as prose in `src/lib.rs:17` and the README. Exit mapping is `Clean|Advisory→0`, `Failing→1` (`src/finding.rs:121-126`) — no 2/3/4, and `ValidatorError` (`src/error.rs:39-46`) has no exit code at all, so a refusal has no defined process code and nothing distinguishes "non-zero exit, valid payload" from "non-zero exit, no payload".

None of that is a rule violation — nothing bare-function-shaped is exposed over IPC, because nothing is exposed over IPC. **A library crate is an acceptable decomposition and I am not asking for the binary in this PR**, since `quoin-core` lives in #375 and that branch has not merged.

But #377 was chosen for one reason: it is 164 lines behind one command, deliberately trivial, so that the IPC boundary, the exit taxonomy, the generated types and the delete step get proved on a domain with no surprises. If it merges as a library and the ticket closes, none of that is proved, and the first real integration happens in Stage 3 on `semantic` — the exact blast radius validators exists to avoid.

So: **#377 does not close when this merges.** The cutover — registering `validators.*` in `quoin-core` dispatch, the exit taxonomy including the non-zero-but-valid-payload case, `schemars`-generated hash-asserted TS types, and the TS caller switched over reversibly — is a named follow-up that must land before Stage 3 begins. I am filing it and linking it here.

kreneskyp:
Follow-up filed as #412 — Stage 2 cutover: wire `quoin-validators` through `quoin-core` dispatch, full 0/1/2/3/4 exit taxonomy including the non-zero-but-valid-payload case, `schemars`-generated hash-asserted TypeScript types, reversible cutover, then delete code and tests together. It gates Stage 3's cutover.

#377 does not close when this PR merges; it closes when #412 does.

Also: #401 merged to `main` as `df70470`, so the workspace, the 1.98.1 pin and a Rust CI lane now exist. Rebase onto it and drop this branch's private workspace — `members = ["crates/*"]` picks the crate up automatically. Golden and provenance conventions are being consolidated under #411.

kreneskyp:
Fixes pushed as `3f35de4`. Both blocking findings closed, and the divergence resolved rather than recorded. Verified the tree myself.

## The TS lane

`vite.config.ts` now carries `"rust/**"` in `exclude`, with a reason written in the style of the entries beside it. The capture moved out of the test namespace entirely — `tests/golden/generate-oracle.test.ts` is gone, replaced by `tools/generate-oracle.mts` with its own `tools/vitest.oracle.config.mts`. Both halves, as asked: the exclusion stops vitest collecting the tree, and the rename means the capture is not a test regardless of any collection config. Its broken `../src/commands/validate.js` import is fixed and it reproduces the committed `expected.json` byte-for-byte.

Collection is **108 files on this branch and 108 on `df70470`**, which is the number that closes the finding.

The remaining failure is `tests/bench-tier1.test.ts`, failing identically on main: `ENOENT corpus/config/metrics.json` and `python3: can't open file corpus/bounds.py` — the `qa-corpus` submodule is not checked out. Environmental and pre-existing, and confirmed against main rather than assumed.

## The eager-vs-lazy divergence — fixed, not recorded

Direction: **the Rust was eager, the TypeScript is lazy.** `inspectEmptyGates` calls `readFileSync` inside `wiring.find(...)`, so a repo whose scripts declare no negative gate claim opens *no* wiring file, and a script matched by the first candidate never opens the rest. The port read every wiring file up front.

That is not a performance difference. An unreadable `Makefile` the oracle never opens turned a clean verdict into a `QV-E003` refusal — a verdict flip on a repository the oracle passes.

**Matching the oracle is right and recording it would have been wrong.** A divergence entry is for behaviour we are choosing to keep; this was strictly worse behaviour that no user wants and that cost nothing to fix. Recording it would have preserved a defect under a heading — which is the failure mode this programme keeps hitting.

The implementation is lazy-with-memoisation in `src/gates.rs`. The parity argument is the part that matters: the **set of files opened** is now byte-identical to the oracle's, and the cache only removes re-reads of files already opened, which cannot change that set. `tc_377_023` pins both halves — unreadable-but-untouched yields clean, unreadable-and-needed yields `QV-E003` naming the path. PROVENANCE records the cache as an explicit non-divergence, which is the right place for it.

## Gate

`fmt --check` clean; `clippy --all-targets --all-features -- -D warnings` clean; 51 tests pass across `quoin_core` lib (21), `tc_boundary` (6), `tc_source_conventions` (3), `quoin_validators` lib (15), `golden_parity` (5, `tc_377_019`–`tc_377_023`) plus a doctest; `advisories ok, bans ok, licenses ok, sources ok`; `cargo metadata` shows `quoin-validators` resolving as a workspace member of the merged root.

## Two corrections to my review instructions, both worth carrying

**`cargo deny` flag order.** `cargo deny check --manifest-path …` is rejected; the working forms are `cargo deny --manifest-path rust/Cargo.toml check` or the repo's own `cd rust && cargo deny check` (`Makefile:297`). This is the second branch to hit it, so it is my instruction that is wrong, not a branch quirk.

**`--reporter=basic` no longer exists in vitest 4.1.10** — it is resolved as a custom reporter module and fails with `ERR_LOAD_URL`. True on main as well. `--reporter=dot` is the replacement.

## One addition outside the brief, accepted

`tests/golden/expected.json` added to `.prettierignore` with a stated reason, in the style of the existing `src/quire/schemas/` entry. `prettier --check .` covers `rust/**` and wanted to re-indent the golden, which would break the byte-reproduction check. `cases.json` beside it is hand-authored input and is left formatted. Correct call — a formatter rewriting a byte-compared artefact is the same class of problem as a capture script rewriting its own golden, which is what this PR started out fixing.

Scope fences held: no Trace retagging (#410), no goldens-layout change (#411), no workflow file, no pin weakened, no lint relaxed.

Standing: #377 does not close when this merges. It closes when #412 does.

Issue #380, opened 2026-09-12
[Quoin#380] [Stage 4 foundation] quoin-store: canonical JSON, JCS, blake3, atomic rename


kreneskyp:
Build `rust/quoin-store`: canonical JSON serialization, RFC 8785 JCS, blake3 digests, atomic rename.

**The highest-risk crate in the port.** Digests are *identifiers* inside retained evidence records, so a mismatch is not a failed test — it is unrecoverable corruption of existing evidence.

## Hard gate before any cutover depends on this
Replay every digest in every reachable store through both the TS and Rust implementations. **Zero mismatches or no cutover.**

Test JCS canonicalization separately and first, adversarially — unicode normalization, number formatting, key ordering. That is the real risk surface; blake3 itself is a well-specified function and is not where this breaks.

## Acceptance
- one digest function in one place, used by everything downstream
- `STORE_SCHEMA_VERSION` and the canonical-JSON serialization treated as a frozen compatibility surface
- every store in the ecosystem: read -> re-serialize -> byte-identical
- adversarial JCS suite committed

---
Parent: #373. Depends on the workspace from #375 for wiring, but the crate itself is written and tested standalone — do not block on it.

Standing rules: `/rust-review` is the style guide, a repo `rust-style` doc outranks it. `#![forbid(unsafe_code)]`, one `thiserror` error enum per crate boundary with stable codes, newtypes over bare `String`/`Uuid` for ids. The existing TypeScript is the oracle: capture expected fixtures from it once, never leave TS as a runtime oracle. **No TypeScript is deleted in this ticket** — cutover and deletion are separate tickets per FR-018.

PR #401, opened 2026-09-12
Stage 0: the Rust workspace, the boundary and one candidate revision (#375)

### LINKED ISSUE #375: [Stage 0] Rust workspace, boundary contract and CI for the burn-down

kreneskyp:
Closes #375. Stage 0 of the burn-down — the stage that freezes TypeScript growth before anything moves.

Lands `rust/` as a single in-repo Cargo workspace beside the retained `src/`, pinned at **1.98.1**, with `[workspace.lints]` and the CI leg that runs it.

Not a second repository: FR-018's "one candidate revision" rule makes a separate repo a pin-and-sync drift problem — the exact class `quire-cli` hit when it shipped five tags each reporting the previous version. `engineering-assurance` already demonstrates in-repo Rust beside retained legacy.

CI is `on: workflow_dispatch:` only, per standing policy — hosted runners cannot install `@agent-ix/*` from npm.ix.

## Open items this ticket owns, recorded in comments

- Wire #379's `quoin-quire` and #381's `quoin-config`/`quoin-modules` as members and **delete their standalone `[workspace]` tables and per-crate `Cargo.lock`s**. A crate that keeps its own workspace table still builds, but sits outside `[workspace.lints]`, so `cargo clippy --workspace` goes green because the crate was never in scope.
- `allow-expect-in-tests`.
- Run `cargo deny check` once `quoin-quire` joins, and rule on `serde_yaml_ng` → `unsafe-libyaml` against the crates' `#![forbid(unsafe_code)]`.


kreneskyp:
Reviewed. **MERGEABLE WITH FIXES** — five, all in workspace configuration, none in code. Gates were re-run rather than read: `cargo +1.98.1 fmt --all --check` clean, `clippy --workspace --all-targets --locked -- -D warnings` clean, `cargo test --workspace --locked` green. Exact pins are centralized in `[workspace.dependencies]`, `publish = false` is present on all three crates, and the toolchain triple-pin test, error taxonomy, canonical-JSON discipline and Makefile target-dir hygiene are sound.

The fixes exist because this branch is the base under five parallel branches (#376, #377, #378, #379, #380). Anything here that a sibling must edit is a merge conflict five times over.

### Required before merge

1. **`rust/Cargo.toml:16-20` — `members` is an explicit three-entry list.** Every sibling branch must edit the same five lines. Use `members = ["crates/*"]`.

   The membership list is also unenforced, which is worse than the conflict. `tc_375_every_crate_opts_into_the_workspace_lint_policy` (`rust/crates/quoin-core/tests/tc_source_conventions.rs:74`) walks `crates/` **on disk**, and nothing asserts that every directory it finds is a workspace member. A branch that adds a crate and forgets the `members` edit still builds, still passes tests, and sits silently outside `clippy --workspace`. The PR body warns about this for #379 and #381 without guarding against it. The glob removes the failure mode rather than testing for it.

2. **`rust/deny.toml:25-28` — per-crate AGPL `exceptions` is the same five-way edit on a second file.** `[licenses] private = { ignore = true }` covers every `publish = false` crate permanently and deletes the list.

3. **`rust/Cargo.toml:56-61` — no `jsonschema` entry.** The pin is already ruled: `=0.56.0, default-features = false`. It is recorded in the plan and nowhere in the workspace, so #377 and #379 will each choose one independently. Add it now, unused, exactly as the other pins sit. The argument for that pin is that "no network on a command path" becomes a property of the dependency graph — which holds only while the pin is central and exact.

4. **`rust/Cargo.toml:47-54` — `[workspace.package]` omits `publish = false`.** Each crate restates it by hand, and unlike the lint opt-in no test asserts it. One `publish = false` in `[workspace.package]` plus `publish.workspace = true` per crate makes it structural. We are not publishing to crates.io, and that should be impossible by construction rather than by three correct copies.

### Rulings on the remaining question

**The narrow licence allow-list at `rust/deny.toml:17-20` stays as it is.** The reviewer flagged it as a Stage 0 decision scheduled for re-litigation because each new dependency must widen it. That is the intended behaviour: a new licence entering the graph should require a decision, and a list that already admits everything enforces nothing. Widening it per dependency is the process working.

**`multiple-versions = "deny"` at `rust/deny.toml:33` also stays, and the header comment pre-conceding its relaxation should be deleted.** A branch that hits a duplicate version resolves the duplicate; it does not relax the gate. Whichever sibling lands first would otherwise set the precedent for all five, and lowering a gate to land a branch is how the gate stops meaning anything. If a duplicate turns out to be genuinely unresolvable, raise it and it gets decided on its own merits — not inside an unrelated PR.

Fixes 1-4 are mechanical. Once they land, the five sibling branches can merge without touching a shared line, and each drops its private `Cargo.lock`.

PR #402, opened 2026-09-12
quoin-semantic and quoin-completeness (#378)

### LINKED ISSUE #378: [Quoin#378] [Stage 3] Port semantic and completeness to Rust crates

kreneskyp:
Closes #378. Stage 3 — the leaves with no dependents, chosen so the ajv↔`jsonschema` error-shape gap gets discovered on a small blast radius instead of inside measurement.

`rust/crates/quoin-semantic/` and `rust/crates/quoin-completeness/`, ~6,900 lines, `#![forbid(unsafe_code)]`, one `thiserror` enum per crate with stable codes (`QSEM-001..008`, `QCMP-001..002`), newtypes for every id, no `anyhow` and no `Box<dyn Error>`.

The crates **read `src/semantic/schemas/` rather than copying it**, so the two implementations cannot drift onto different bytes.

## Verdict parity is exact: 171/171

On both `jsonschema` 0.56.0 and 0.18.3. No TS bug and no spec gap surfaced. That is the contractual half of the Stage 3 risk closed.

## Four diagnostic divergences, all absorbed (D4 was not one of them)

Full report at `rust/crates/quoin-semantic/DIVERGENCE.md`. **D4 is now closed** — see below.

- **D1 `additionalProperties`** — ajv emits one error per key, `jsonschema` one error listing all. Absorbed because a manifest with two typos would otherwise report one and the user round-trips. The adapter fans out. 1/118 docs.
- **D2 `anyOf`/`oneOf`** — ajv reports failing branches alongside the composite, `jsonschema` nests them under `context`. Absorbed because an unknown `targets` value would otherwise lose the allowed-values list *and* stop selecting `semantic.unknown-target`, degrading to `semantic.invalid-value`. The adapter flattens. 3/118 docs.
- **D3 order is not contractual and cannot be** — 110/118 match ajv, and the two `jsonschema` releases disagree with *each other* on `exports-two-faults`. The port sorts on `(instance location, keyword)`.
- **D5 map iteration order** in `resolveImports` — result set identical, cycle-trail order can differ.

Explicitly measured as **non**-divergent: lookahead patterns (both use `fancy-regex`, so `common.schema.json`'s `sourceLocus.path` compiles), `unevaluatedProperties`, `prefixItems`, `dependentRequired`, integer-valued floats, `oneOf`+`not`, `required` multiplicity.

## The `jsonschema` pin: `=0.56.0`, `default-features = false`

The ticket described a two-way ecosystem split; it is really 1-vs-4. quire-rs pins `~0.18`; quire-contract-ir, quire-spec-language, quire-contract-codegen and quire-analyze all pin `=0.17.1`; filament-core-data pins none.

1. **0.18.3 is measurably non-conformant on 2020-12** — it accepts `[1]` against a `$dynamicRef`→string schema that both ajv and the spec reject. No vendored schema uses `$dynamicRef` today, but filament-core-service and filament-core-data own those schemas, not quoin.
2. Nothing links quoin's crates to quire-rs or quire-contract-ir. #373's boundary is a subprocess, so there is no unification constraint to satisfy.
3. `default-features = false` drops `resolve-http`/`resolve-file`, which makes "no network on a command path" a property of the dependency graph rather than a convention. 0.18's *default* resolver reads files and HTTP.
4. `=` rather than `~` because the error-kind match is exhaustive with no `_` arm, so a version bump is a reviewed change.

## Goldens

171 schema documents plus behavioural goldens (26 `readSemanticBlock` cases, 28 classifier docs, a 24-cell verdict grid, 20 written-reason cases, end-to-end `assessBundle`), all captured once by `scripts/capture-semantic-goldens.mjs` from quoin@`4d27dcf`, ajv 8.20.0, recorded in each file's `provenance` block and asserted by a test. Regeneration is byte-stable. **Nothing in the Rust test lane invokes Node** — no non-Rust runtime oracle survives the cutover, per FR-018.

Parity assertions are mutation-checked: collapsing the D1 fan-out fails `tc_378_100`; relaxing `MIN_REASON_WORDS` 3→2 fails `tc_378_300`.

## Gate

`cargo +1.98.1 fmt --check` and `clippy --all-targets --all-features -- -D warnings` clean on both crates. 77 Rust tests passing, 0 failed. `prettier --check .` clean. The five TS suites over the ported areas still pass (46 tests) — nothing is deleted here.

## D4 — YAML scalar resolution: a verdict-level defect, now closed

D4 was recorded here and in `DIVERGENCE.md` as diagnostic-only with zero
measured impact. Both halves were false, and both are fixed in this PR.

**The premise was falsified inside this PR.** `bundle.rs` read frontmatter with
no schema validation at all, and `strings_at` silently *dropped* any scalar that
did not resolve to a string. So a token the two readers resolve differently did
not change a diagnostic — it changed the **verdict**.

**"Measures zero" was measured over an empty population.** The golden corpus
contained no ambiguous token, so the check had nothing to check.

**Reproduced before the fix** (corpus extended, no code change yet):

```
---- tc_378_303_bundle_frontmatter_matches_the_oracle stdout ----
assertion `left == right` failed: projected claims
  left: [..., ("ambiguous/claims.md", ["security", "on", "no", "1:30", "2026-09-12", "0b101"], []), ...]
 right: [..., ("ambiguous/claims.md", ["security", "on", "no", "1:30", "2026-09-12", "017", "010"], []), ("ambiguous/excuse.md", [], ["017"]), ...]

---- tc_378_304_assess_bundle_matches_the_oracle_end_to_end stdout ----
assertion `left == right` failed: advisory: verdict
  left: "PASS"
 right: "FAIL"

test result: FAILED. 4 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

After: `test result: ok. 6 passed; 0 failed`.

### Correction: the reported tokens are not the divergent ones

#404 and the old `DIVERGENCE.md` table claim `12:30`, `yes`/`no`/`on`/`off` and
dates diverge. Measured against `serde_norway` 0.9.42, they do **not** — that
reader already resolves all of them as strings, exactly as the oracle does. The
old table was wrong about the library it described.

The tokens that actually diverged, measured:

| Frontmatter | TS oracle (YAML 1.2 core) | `serde_norway` (YAML 1.1) |
|---|---|---|
| `017` | `17` | `"017"` |
| `010` | `10` | `"010"` |
| `0b101` | `"0b101"` | `5` |

### How it is closed

Fix shape **(i)**: a new workspace crate `quoin-yaml` wrapping `saphyr`
(`=0.0.12`), a native YAML 1.2 core-schema reader, so resolution matches the
oracle by construction. Shape (ii) — coercing scalars back to their source text
— would match only for string-typed fields and would *move* rather than close
the divergence for genuinely numeric ones.

A new crate rather than a reader duplicated in two places or a dependency edge
between two sibling leaf crates.

The corpus now carries `on`, `no`, `null`, `1:30`, `2026-09-12`, `017`, `010`
and `0b101` as module and keyword values, with verdict parity asserted against
the oracle on each. The non-diverging tokens stay in as regression cover.

Residual divergences, all listed in `DIVERGENCE.md` §6 with their user-visible
consequence: parse-error message wording, `.inf`/`.nan` → `null` (matching what
the oracle's own `JSON.stringify` writes), and duplicate mapping keys.

#404 stays open — its Stage 4 scope (free-valued frontmatter reaching the
evidence store and the digest-replay gate) is not addressed here.

## Workspace join (#375)

Both crates now live under `rust/crates/` as members of the root workspace.
Private `[workspace]` stanzas and lockfiles deleted; package metadata, lints and
shared dependencies inherited. No per-branch workflow file added.

Two root-level changes were forced by the join:

- **`thiserror` `=2.0.18` → `=2.0.20`.** `saphyr` requires `^2.0.20` and
  `multiple-versions = "deny"` forbids carrying both. A raise, not a relaxation.
  `feat/381-config-modules` makes the same bump independently, so the branches
  converge rather than conflict.
- **`serde` `=1.0.228` → `=1.0.229`.** Moves `serde_derive` onto syn 3 — see the
  `cargo deny` section below.
- **`MIT-0` and `Zlib` added to `deny.toml`'s allow-list**, per that file's own
  stated convention. Both arrive with `jsonschema` as it enters deny scope for
  the first time: `MIT-0` via `borrow-or-share ← fluent-uri ← referencing`,
  `Zlib` via `foldhash ← hashbrown ← referencing`.

### `cargo deny`: the `syn` duplicate, resolved on our side and skipped on theirs

`bans` failed on `error[duplicate]: found 2 duplicate entries for crate 'syn'`
(2.0.119 and 3.0.5). I stopped rather than skipping, and the ruling came back:
skip it, because the duplicate is not ours.

**Measured provenance disagreed with that premise in part.** `cargo tree -i
syn@2.0.119` put `serde_derive 1.0.228` on the syn 2 side, reaching our crates
directly — not only `strum_macros` under `jsonschema`. Skipping with that edge
live would have conceded a duplicate that was half ours.

So the half we own is **resolved**, not skipped: `serde` raised `=1.0.228` →
`=1.0.229`, whose `serde_derive` builds on syn 3, matching `thiserror-impl`
2.0.20. After that raise:

```
syn v2.0.119
└── strum_macros v0.28.0 → strum v0.28.0 → jsonschema v0.56.0 → quoin-semantic
```

That is the case `[[bans.skip]]` exists for, and it is the only entry in the
file. It is version-exact on `2.0.119`, leaving `3.0.5` counted — 3.0.5 is the
side our own pins land on, so it stays under the gate, and any *other* `syn`
pair still fails.

The `reason` in `deny.toml` carries the full chain for each side, that
`jsonschema` 0.56.0 is the latest release, that `strum_macros` has no syn 3
release, and **how that last claim was checked** (`cargo info strum_macros` →
0.28.0 latest; `[dependencies.syn] version = "2.0"` in its manifest) so a future
reader can re-verify it — it goes stale the moment `strum_macros` ships one.

Contrast recorded in the file, from the ruling: on `feat/381-config-modules`
this same `syn` pair is OURS and is resolved there by raising both pins, not
skipped. Same crate, opposite ruling, because the provenance differs.

`advisories ok, bans ok, licenses ok, sources ok`.

## `serde_json/preserve_order` (new `DIVERGENCE.md` §10)

Measured, not assumed: enabling it anywhere in the workspace breaks exactly
`quoin-core::protocol::tests::canonical_json_sorts_keys_and_emits_no_whitespace`
and `tc_375_stdout_is_canonical_json_one_line`. Feature unification makes it a
workspace-level hazard, not a crate choice. Noted on #375.

## Gates

| Gate | Result |
|---|---|
| `cargo +1.98.1 fmt --manifest-path rust/Cargo.toml --all --check` | clean, no output |
| `cargo +1.98.1 clippy --manifest-path rust/Cargo.toml --all-targets --all-features --locked -- -D warnings` | clean — `Finished dev profile [unoptimized + debuginfo] target(s) in 0.43s` |
| `cargo +1.98.1 test --manifest-path rust/Cargo.toml --locked` | all green — completeness lib 17 + golden_parity 6, core lib 21 + tc_boundary 6 + tc_source_conventions 3, semantic lib 39 + golden_parity 15, **quoin-yaml lib 4** |
| `cargo +1.98.1 deny --manifest-path rust/Cargo.toml check` | `advisories ok, bans ok, licenses ok, sources ok` |
| `cargo metadata --no-deps` | 6 workspace members; both crates at `rust/crates/` |

`DIVERGENCE.md` §9's reproduction command had **no** `--manifest-path` at all
(not a wrong one), so it could not run from the repository root. Corrected and
verified.

## For the integrator (#375)

- No `rust-toolchain.toml` and no per-crate `Cargo.lock` — both belong to the workspace root.
- `quoin-completeness` deliberately does not port `defaultModuleRoots()` (catalog, Stage 7); the caller supplies roots. `locateModuleRoot` is duplicated with a named successor, per NFR-024.


kreneskyp:
Reviewed. **BLOCKED** — one of the five divergences is not a divergence.

First, what is right, because it is most of the PR. The corpus is real and non-empty: 65/47/59 documents, 53 valid, 118 invalid, zero invalid-without-errors. ajv is the oracle, and each case's `expect` cross-checks it (`scripts/capture-semantic-goldens.mjs:155-158`), so no golden is a fixture the test wrote for itself. The tests assert population floors and that every finding kind and every verdict actually occurs, which is the guard against a suite that passes because it measures nothing. And no YAML round-trip work was taken on — `serde_norway` is `from_str` only, with no serializer anywhere, which is correct: quoin never round-trips YAML and this PR does not start.

Adjudicating the five:

- **D1 `additionalProperties`** — real and closed. `jsonschema` returns `unexpected: Vec<String>` and the adapter fans it out at `src/schema.rs:320-330`. Library-level, not fixable upstream.
- **D2 `anyOf`/`oneOf`** — real and closed. `context` nesting is structural in the crate; `push_error` recurses (`src/schema.rs:358-376`) and identity multisets then match on all three affected documents. The §4 caveat about deeper composites is honest and stays.
- **D3 order** — real and unclosable, handled correctly. Two `jsonschema` releases disagree with *each other*, so there is no stable order to match; the port sorts on (path, keyword) and the goldens compare multisets, never order. This is precisely the "verdicts contractual, error text not" line, applied.
- **D5 map order** — real, message-only. §7's "the set of cycles found is identical" is asserted rather than proved — a DFS with a `done` memo can report a different cycle count under a different visit order — but nine golden cases measure it, so the residual risk is small. Leave it; state it as asserted.

### BLOCKING — D4 is a live verdict-level defect in this PR, not a Stage 4 deferral

`DIVERGENCE.md:263` bounds D4's blast radius with "the schema refuses a non-string". That premise does not hold inside this PR. `rust/quoin-completeness/src/bundle.rs:16-20` says the frontmatter reader does no schema validation, and `strings_at` (`bundle.rs:255-263`) silently **drops** non-strings.

So under YAML 1.1 a vocabulary value of `no`, `on`, `y`, or `017` parses as a bool or an int, gets dropped, and the value reports as `unowned` where the TypeScript records it as claimed. That is a verdict flip in this crate's own product, today — not a diagnostic difference, and not something Stage 4 inherits.

Fix: read frontmatter with a YAML 1.2 reader, or coerce scalars back to their source text. Either way, add corpus cases.

**And the claim that D4 measures zero has no measurement behind it.** `tests/goldens/bundle-frontmatter.json` contains no YAML 1.1/1.2-ambiguous token at all — no `no`, `on`, `12:30`, `017`, `1.10`. §0 of this document says it does not do that. A bound asserted over a population that contains none of the thing being bounded is the empty-population case, which is the one failure mode FR-018 names by hand.

Compounding it: **the PR description omits D4 entirely.** It lists D1, D2, D3, D5 and says "four divergences, all in diagnostics". A reviewer reading only the description sees a value-level divergence described as diagnostic-level. Fix the description in the same push.

### Must also fix before merge

**The parity suite is not gated.** `.github/workflows/build-test.yml` runs `cargo build` only for `.ci/quire-cli`; nothing builds, lints or tests `rust/`. The entire deliverable of this PR runs only on the author's machine. This is the same finding I raised on #398 — every Stage 1–3 branch is shipping its gate into a lane that does not exist. Add it (`workflow_dispatch:` only, per standing policy).

**The documented re-measure procedure does not run.** `DIVERGENCE.md:334-335` gives `cargo test -p quoin-semantic`, but this branch has no `rust/Cargo.toml`, so `--manifest-path` is required. A reproduction procedure that fails when followed is not a reproduction procedure.

### Follow-up

`quoin-semantic/Cargo.toml:29` enables `serde_json/preserve_order`; `quoin-completeness` does not. Feature unification is a workspace property, so when #375's root manifest lands, completeness's `serde_json::Map` flips from sorted to insertion order with no code change in its own crate. Note it on the #375 integration ticket so it is caught deliberately rather than as a golden diff nobody can explain.

kreneskyp:

kreneskyp:
## Fix verification — all items closed, `fca01d1`

| Gate | Result |
|---|---|
| `cargo +1.98.1 fmt --manifest-path rust/Cargo.toml --all --check` | exit 0 |
| `cargo +1.98.1 clippy --all-targets --all-features --locked -- -D warnings` | clean |
| `cargo +1.98.1 test --manifest-path rust/Cargo.toml --locked` | completeness 17 + 6, core 21 + 6 + 3, semantic 39 + 15, quoin-yaml 4 — zero failures |
| `cargo +1.98.1 deny --manifest-path rust/Cargo.toml check` | **advisories ok, bans ok, licenses ok, sources ok** |
| `cargo metadata --no-deps` | 6 members, both crates under `rust/crates/` |

### The `syn` duplicate, and a correction to my own ruling

I ruled that both `syn` edges sat inside `jsonschema`'s subtree and that the pair was therefore unresolvable and skippable. **That was wrong, and `cargo tree -i syn@2.0.119` said so:**

```
syn v2.0.119
├── serde_derive v1.0.228 → serde v1.0.228 → quoin-core, quoin-completeness, quoin-semantic
└── strum_macros v0.28.0 → strum → jsonschema v0.56.0
```

`serde_derive` reached our crates directly through our own root pin. Writing the skip with that edge live would have recorded a false reason and conceded a duplicate that was half ours.

Our half was resolved first — `serde` `=1.0.228` → `=1.0.229`, whose `serde_derive` builds on `syn 3` (confirmed in the vendored manifest, `[dependencies.syn] version = "3"`), matching `thiserror-impl 2.0.20`. Only then is the tree what the ruling described:

```
syn v2.0.119
└── strum_macros v0.28.0 → strum v0.28.0 → jsonschema v0.56.0 → quoin-semantic
```

and only then is `2.0.119` skipped, version-exact, with `3.0.5` left counted.

This is the same treatment `feat/381-config-modules` gives the same pair, and that contrast is written into `deny.toml` itself: **same crate, opposite ruling, provenance is the test.** `deny.toml`'s "a duplicate version is resolved, not conceded" means we resolve what we control; `strum_macros` under `jsonschema` is not something we control.

The `reason` records the method behind its most perishable claim — `cargo info strum_macros` shows 0.28.0 is latest, and `[dependencies.syn] version = "2.0"` in the vendored source — with the instruction to **drop the entry rather than bump it** when `strum_macros` ships on `syn 3`.

### D4 — closed

New workspace crate `rust/crates/quoin-yaml` wrapping `saphyr =0.0.12`, a YAML 1.2 core-schema reader. Fix shape (ii), source-text coercion, was rejected **in writing**: it matches the oracle only for string-typed fields and would move the divergence rather than close it for genuinely numeric ones.

Reproduced both directions — `advisory: verdict` `left: "PASS"` / `right: "FAIL"`, 2 failing → 6 passing.

### The divergence table was factually wrong, and is now measured

`serde_norway 0.9.42` already resolves `on`/`off`/`yes`/`no`/`y`/`n`/`1:30`/`12:30:45`/`2026-09-12` as strings and `null`/`~` as null — **matching the oracle exactly**. Both `DIVERGENCE.md`'s old table and #404's table were wrong about the library actually in the tree.

The measured divergent tokens are `017` (oracle `17` / libyaml `"017"`), `010` (oracle `10` / `"010"`), and `0b101` (oracle `"0b101"` / `5`). #404 is corrected ([issuecomment-5648689845](https://github.com/agent-ix/quoin/issues/404#issuecomment-5648689845)) so Stage 4's corpus is not seeded with inputs that do not diverge — a corpus of non-divergent inputs would read green forever and prove nothing.

### Other items

Workspace join complete: both crates under `rust/crates/`, private `[workspace]` stanzas and lockfiles deleted, everything inherited, non-root deps `=`-pinned. §6 of `DIVERGENCE.md` rewritten with the falsified premise removed rather than annotated. New §10 on `preserve_order`, §11 recording both pin raises with the note that `feat/381-config-modules` raises `thiserror` to the same `=2.0.20` independently, so whoever merges second does not read it as drift.

`thiserror` `=2.0.18` → `=2.0.20` is a raise (saphyr needs `^2.0.20`), not a weakening.

### Standing correction

`cargo deny check --manifest-path …` is not a valid invocation — the flag precedes the subcommand: `cargo +1.98.1 deny --manifest-path rust/Cargo.toml check`. `DIVERGENCE.md:334-335` had no `--manifest-path` at all and could not run from the repo root under any value; fixed and verified.

Issue #413, opened 2026-09-12
[Quoin#413] The evidence store is the coverage record, and it has never been fed


kreneskyp:
Design decision and its consequences, taken 2026-09-12. Touches FR-030 (evidence store) and FR-032 (auditor). Found while normalising criterion trace tags under #390.

## The decision

**The evidence store is this repository's coverage record. Criterion tags in test source are its input, not a parallel system.**

The reason is not that the store has more machinery, though it does — auditor, ratchet, baseline, affirmation path, commit pin, all built. It is **what the two mechanisms can assert**:

- A **source tag** records a *claim*: somebody wrote `// Trace: FR-030-AC-10` above a test.
- A **binding** records an *execution*: it exists because a test ran and emitted a trace property.

The obligation the Rust burn-down needs is that a criterion never goes silently unbound when its TypeScript test is deleted at cutover. That is a statement about a test **that ran**. A source-tag scan reads green over an `#[ignore]`d Rust test, over a `describe.skip`, and over a suite nobody executes — for exactly as long as nobody runs anything. It cannot carry a cutover. A binding can.

So the binding path is not vestigial. **It was never fed.**

## The three facts this rests on, each verifiable in seconds

1. **`spec/evidence/bindings.json` is empty** — literally `{"bindings": [], "schemaVersion": 1}`. Zero bindings.
2. **Therefore all 341 baseline entries are `undischarged:` by construction.** A `Counter` over `spec/evidence/baseline.json` `accepted` gives `{'undischarged': 341}` and no other kind. It is not a curated list of accepted hard cases; it is the entire enumerated obligation population with nothing bound to anything.
3. **Nothing emits the input.** `src/evidence/adapters/junit.ts:65` reads trace ids from `<testcase>` **properties** (`name="trace"` / `"traceIds"` / `"trace_ids"`), not from test names. Quoin's own vitest run emits none: `grep -c 'property name="trace' reports/junit.xml` → 0.

#390 has just made criterion tags consistent across the test tree — 669 criteria bound by a tag, one tag line per test, one vocabulary. That is exactly the input the store has never had.

## What must happen, in this order

The order is not arbitrary. Each step arms a hazard in the next.

### 1. CON criteria are not enumerated as obligations — prerequisite

The suffix census over the 341 is `{'AC': 317, 'M': 24}`. **Zero CON.** In a set that *is* the whole obligation population, that means the deriver never mints them.

[...]

Issue #410, opened 2026-09-12
[Quoin#410] Merge the burn-down spec artifacts, rebase the port branches, and tag the port tests


kreneskyp:
## Rewritten — the original premise was wrong

This issue was opened on the belief that the five port branches carry Trace tags naming criteria that do not exist (FR-096..FR-100, NFR-024, NFR-025), and that the fix was to author the spec artifacts. **Neither half holds.** The premise was mine and I am recording the correction rather than editing it away.

A tag scan across all five branches (`fix/376-break-dependency-cycles`, `feat/377-quoin-validators`, `feat/378-semantic-completeness`, `feat/379-quoin-quire`, `feat/381-config-modules`) returns **zero unresolved criterion ids**. Not one FR-096..FR-100 or NFR-024/025 tag exists on any of them.

The documents exist too. `git ls-tree origin/spec/373-rust-burn-down spec/functional/` returns FR-096 through FR-103; the same command against `origin/main` returns nothing. They are unmerged, not unwritten. FR-096 alone declares 11 criteria.

The original sightings came from three places, none of them a Trace tag:

- **prose docstrings** — `src/core/index.ts:2` reads "The quoin↔quoin-core boundary (quoin#373 FR-096)". A comment, correctly not flagged.
- **fixture data** — `tests/battletest.test.ts:116` and `bench-corpora.test.ts:432` contain the path `spec/backend/functional/FR-097-session-binding.md`. That is a *different repository's* FR-097 inside our test corpus: same id, unrelated requirement. A grep finds it; a tag check does not, and should not.
- **the documents' absence from `main`**, which makes them look unauthored from a main-based branch.

## What the scan did find

**62 unresolved tags per branch, identical across all five** — they share a pre-migration base and none of them touches the failing files. Every one is a Trace line naming a TC id where a criterion belongs:

```
// Trace: FR-070-CON-2
// Trace: TC-1343
it("records filament-core-service provenance for the module-manifest schema...")
```

That is the pre-migration tag shape. #390's migration converts exactly this, which is why a tree carrying that work reports 0. These resolve on rebase — no authoring, no tagging, nothing to decide.

## The real finding

**The new port tests carry no criterion tags at all.** `tests/core-exec.test.ts` and `tests/core-exec-e2e.test.ts` on `origin/main` both return zero matches for `// Trace:`. Not a wrong tag — no tag. They name FR-096 in a docstring and bind nothing.

[...]

Issue #390, opened 2026-09-12
[Quoin#390] One binding between a test and what it discharges: the criterion id


kreneskyp:
**Status: landed on `spec/373-rust-burn-down` (#414).** This is the complete account, rewritten in place — the earlier text carried a false claim and two appended corrections, and a reader should not have to reconstruct the conclusion from them.

## What was wrong

A criterion could be bound to its test three different ways, all in live use:

1. **Criterion tag in the test** — `// Trace: FR-034-AC-9`. The majority.
2. **TC id in the test**, joined to the criterion through the TC table in `spec/matrix.md` — sometimes both (`// Trace: FR-034-AC-19, TC-195`), sometimes a bare `// TC-1531`, sometimes only a `describe()` **range**, `"TC-1530..1536 …"`.
3. **Prose in the matrix row** — `CON-1/CON-2 → inspection of src/evidence/adapters/`, with no tag anywhere.

One file mixed them. `tests/finding-record.test.ts` tagged AC-1..AC-15 by criterion alone and AC-16..AC-20 by criterion *and* TC id.

**Consequence: every simple reconciliation produced a different wrong answer.** Measured on this repository, each join narrowing the last: 212 absent ids → 10 → 471 → 279 → 15. Five joins, five answers, two orders of magnitude. Two of them were acted on and both were wrong.

## The convention

**Every test names every criterion it discharges, by criterion id, in a `// Trace:` tag. That is the sole machine-readable binding.**

- TC ids are out of test source. The matrix TC table stays as the human index — it is how a person navigates, not how a tool joins.
- `describe()` title ranges stop counting as bindings. `"TC-1530..1536"` binds seven criteria to one block with no way to say which assertion covers which, which is the ambiguity that produced two wrong numbers.
- A criterion not discharged by test says so in its FR document, with the method named.
- **The id pattern enumerates no suffixes.** `requirementOf` (`src/assurance/graph.ts`) is `/^([A-Za-z]+-\d+)/` — the requirement prefix is the whole grammar and the suffix is unconstrained. A suffix allow-list is a standing bet that nobody mints a new kind, and the engine took no such bet.

## What landed

| | |
|---|---:|
| Test files migrated | 58 |
| Binding sites rewritten | 762 |
| Criteria bound by a criterion tag | **444 → 669** |
| `check-trace-tags` | **936 references across 745 tags, 0 unresolved** |

[...]

Issue #388, opened 2026-09-12
[Quoin#388] [Stage 6] Dispose of the corpus-measurement subsystem — do not port


kreneskyp:
**Owner ruling 2026-09-12: never-wired, and it is not a port. Dispose.**

`src/measurement/` contains a second, fully disconnected subsystem — 11 files, 2,117 lines, zero edges to the rest of the measurement domain in either direction, and no importer anywhere.

## Provenance — this is not legacy, it is a closed gate's one-shot harness

- First appears **2026-09-04** in `8cd4159`, *"Corpus measurement over the completed Wave 4 modules (#291) (#351)"*. Eight days old. Touched once since, by `f857454`.
- **#291 is CLOSED**: *"[GATE] Measure the full corpus against completed module schemas (**advisory, report only**)"*. PR #351 merged.
- `grep -rn "measurement/run" src tests scripts evals` returns nothing, and `src/measurement/index.ts` does not export it. `runMeasurement` (`run.ts:108`) and `runDigest` (`run.ts:284`) are unreachable from the CLI.

It never had a caller **by design** — a one-shot advisory harness built for a gate that has since closed, deliberately never given a command.

## What it measured — recorded here, and already retained in the repo

It ran **once**, and its output is committed at `analysis/corpus-measurement/` (9 artifacts, added by `8cd4159`). The capability's *results* survive the code's deletion; nothing needs re-deriving to keep them.

From `run-manifest.json`:

| | |
|---|---|
| Corpus id | `ix-workspace-2026-09-04` |
| Population id | `pop:cafc3ffdf1b73a29` |
| Ran | 2026-09-05T01:48:10Z → 01:48:50Z (**40.0 s**) |
| Repositories enumerated | **251** |

Headline figures (`rates.json`, `states.json`):

| Measure | Value |
|---|---|
| Structural conformance | **7,370 / 7,501 documents = 98.25%** (method `engine-structural-v1`) |
| Properties-form census | **1 / 155 documents = 0.65%** (method `properties-form-census-v1`, 7,432 not-applicable) |
| Document states | measured 7,501 · out-of-model 86 · unreadable 0 · contested 0 |
| Repository health | 3 dirty, 0 unstable |

Retained artifacts, each digest-pinned in `run-manifest.json`: `corpus.json` (2,336 lines), `modules.json`, `states.json`, `census.json`, `rates.json`, `findings.json`, `figure-index.json`, `report.md`, `run-manifest.json`.

[...]

PR #414, opened 2026-09-12
Rust burn-down: spec artifacts, executable-path matrix, and the trace-tag convention

kreneskyp:
Carries the spec artifacts the six open port PRs reference, plus the #390 trace-tag work. 26 commits.

**This unblocks the merge queue.** #397, #398, #399, #400, #402 and #405 are open behind it: their tests reference criteria that exist only on this branch. `git ls-tree origin/main spec/functional/` returns no FR-096..FR-103 today.

## Spec artifacts (AC-2 of #373)

| | |
|---|---|
| **ADR-0003** | Rust-native Quoin engine boundary — options table, capability→disposition, stage order. In `docs/semantic-module-architecture/adr/`, not a second ADR home. |
| **StR-009** | One implementation language for engine logic; one schema source |
| **US-024** | Burn down non-Rust engine logic |
| **FR-096..FR-103** | Versioned boundary · schema-sourced types · semantic/identity parity · Rust catalog+validation · evidence+measurement+change-assurance · retire replaced paths · command surface and oclif retirement · corpus consolidation |
| **NFR-024..NFR-026** | Bounded staged coexistence · immutable evidence and corpus bytes · Rust toolchain floor |
| **IT-003** | Quoin ↔ filament-core-data published types |

## The executable-path matrix (AC-3 of #373)

`docs/rust-burndown/executable-path-matrix.md` plus `.language-allowances.yaml`, the exception manifest the implementation-language policy requires and the first in the ecosystem.

FP-NRX measured, not estimated: **Violations 0 / Retained-with-successor 81,910 / Allowed 23,579**, reconciling to 363 files and 105,489 lines. Zero unclassified rows.

The estimate in #373 was corrected three ways: `skills/` (23,164 lines) was never counted, `corpus/` is a submodule whose lines belong to another repository, and `bin/` plus the root tool configs were missing.

## Disposals

- **#388** — the corpus-measurement subsystem, 3,200 lines across 20 files. A one-shot advisory harness for a gate that closed; `run.ts` had no importer anywhere. Its results are retained at `analysis/corpus-measurement/` and survive the code.
- **FR-084..FR-092 and US-022** withdrawn with banners, matrix rows and a log entry. Documents kept, not deleted — the banner preserves provenance.

## The trace-tag convention (#390)

One machine-readable binding between a test and what it discharges: the criterion id. TC ids are out of test source; the matrix TC table remains the human index.

- 58 files migrated, 762 binding sites rewritten, criteria bound by a criterion tag **444 → 669**
- `scripts/check-trace-tags.mjs`, wired as `make check-trace-tags` — **936 references across 745 tags, 0 unresolved**. Bypass-probed: a planted bad id and a planted empty tag each fail with `file:line`. Refuses an empty population rather than passing over one.
- Six tests corrected from FR-043 to FR-095 — a whole file mis-traced to the wrong requirement, with the matrix disagreeing with itself in both directions

**Not** added to the locked verification stack; that relock is a separate change.

## Reverted in-branch

`b36156d` minted `NFR-<n>-AC-<k>` ids on NFR metric rows before it emerged those obligations already have ids — `NFR-<n>-M-<k>`, used by the auditor's ratchet baseline, documented in `src/assurance/graph.ts` and printed in `evidence affirm`'s own help. Reverted in `0539b21` rather than left pending proof: it added a column to the table those ids are minted from, while they are keys in retained evidence.

## Gates

`make lint` clean · `make check-trace-tags` 0 unresolved · 96 test files, 1124 tests passing.

Refs #373, #388, #390, #410, #413

https://claude.ai/code/session_01J4qKWaqT3xCXMbbtDqdDRY


kreneskyp:
## Review context: dispositions, and two corrections to figures in the body

### SR-159 records all fourteen review findings

`spec/reviews/373-rust-burn-down/dispositions.md` (`7632160`). **Eleven resolved, three open.**

Of the five high findings — three resolved, two open:

| | |
|---|---|
| FND-173 | Resolved *the other way*. It argued `skills/**/workflow-assets/**` should not count as first-party debt; the matrix agrees — `skills/**` is Allowed under the `agent-skills` exception and its 23,164 lines were never in the baseline. |
| FND-175, FND-176 | Resolved by `5968de3`, a commit **on this branch**. |
| **FND-174, FND-177** | **Open**, deferred to [#378](https://github.com/agent-ix/quoin/issues/378). Both are the ajv ↔ `jsonschema` error-shape surface — `format` assertion and `strict`/`allErrors` — and belong to the stage that owns that risk, not here. |

FND-182 stays open with its half-answer stated. The rest are resolved, each with the commit that did it.

**Review files are deliberately not edited.** A review records what was true when it ran; rewriting one to say a finding went away destroys the evidence it was raised. That is also the defect SR-159 fixes: three highs were answered by a commit on this branch and the review still reads them open.

### The FP-NRX figures in the PR body are superseded

The body quotes **81,910 / 23,579 / 105,489 across 363 files**, measured before #388 deleted the corpus-measurement subsystem. Re-measured at `ed5cceb`:

| | Files | Lines |
|---|---:|---:|
| **Violations** | 0 | **0** |
| **Retained-with-successor** | 326 | **78,669** |
| **Allowed** | 18 | **23,579** |
| **Total** | **344** | **102,248** |

**Violations 0 across 344 files is the level. 78,669 across 326 is the slope.** Those are the two numbers the programme reports.

Four things moved it and only two were predicted: `storybook-deploy.js` deleted, #388's 3,200 lines disposed, `check-trace-tags.mjs` added, and **185 redundant `// Trace:` lines merged away** in the tag pass. That last is why the figures are a fresh measurement rather than a subtraction — 81,910 − 3,200 = 78,710 looks right and is wrong by the merge.

### Context for the revert commit a reviewer will hit

`0539b21` reverts `b36156d`, which minted `NFR-<n>-AC-<k>` ids on NFR metric rows. Those obligations already have ids — `NFR-<n>-M-<k>` — and the revert is not a matter of taste:

- `~/.ix/filament/modules/spec-artifacts-process/manifest.yaml` declares an `nfr-metric` obligation source with `id_format: "{document}-M-{row}"`. The engine mints one per Measurement and Evaluation row, in document order.
- `src/assurance/graph.ts` documents the derivation with `NFR-010-M-2` → `NFR-010`; `src/commands/evidence/affirm.ts` uses `--obligation NFR-006-M-2` as its own example; `spec/evidence/baseline.json` carries 24 such keys.

So `NFR-013-AC-1` was a second name for `NFR-013-M-1`, and the reverted change added a column to the very table those ids are minted from while they are keys in retained evidence.

### One thing this branch does not carry

`scripts/check-trace-tags.mjs` is **unwired** — not in the `Makefile`, not in `package.json`. Wiring it changes two digests that `quality/verification-stack-lock.json` pins, so `make test` would fail its own integrity check. The relock is folded into [#350](https://github.com/agent-ix/quoin/issues/350). Run it directly with `node scripts/check-trace-tags.mjs`.

kreneskyp:
## Coordinator review — HOLD, with rulings

Independent review run against `7a1b1c8`. The FR/NFR criteria themselves are the strongest spec artifacts in this repo. Every blocking defect is in the **supporting** artifacts — the allowance manifest, the burn-down matrix, the dispositions document — not in FR-096..FR-103.

Two of the three blockers are the failure mode this set exists to end: **a number that reads green over a population that is not what it claims.**

### Blocking

**B1 — the branch fails its own gate.** `spec/reviews/373-rust-burn-down/dispositions.md` declares `type: SpecReview` but puts findings under `## Dispositions`. `make validate` fails; `make test` depends on `validate` (`Makefile:49`), so the test lane is red on every branch after this one. One-line fix.

(`spec/assurance/AP-201-finding-quality.md`'s duplicate-heading failure pre-exists on `main` and is not a branch regression.)

**B2 — the Allowed class rests on three categories no requirement defines.** `.language-allowances.yaml` declares eight categories. `NFR-024` enumerates five, normatively. Three consequences:

- `NFR-024-AC-8` governs `thin-host`; the manifest spells it `host-dispatch`. **Zero matching entries — the criterion reads green over an empty population.** Same shape as the #405 replay gate reporting PASS with zero comparisons performed.
- `tool-config`, `agent-skills` and `schema-source` appear in no spec artifact and carry **23,527 of the 23,579 Allowed lines (99.8%)**.
- `NFR-024-AC-9` checks allowed→manifest only. Nothing fails a manifest entry whose category the requirement does not admit, so a ninth category can be invented in YAML and every path under it becomes Allowed silently.
- `NFR-024-AC-6` requires an `owner` per entry. No `owner:` key exists in the manifest; it has `authority:` at category level.

"Violations 0" currently depends on which artifact the implementer read.

**B3 — `skills/**` is classified two opposite ways inside one PR.** `FR-099:83` and `FR-101:85` both carry the same normative SHALL: classify `skills/**/workflow-assets/**` as executable assertion logic rather than inert data, because those assets' `specInvariants` decide whether a review, matrix or plan flow passes. ADR-0003 repeats it in bold; StR-009 and US-024 build on it. The manifest exempts `skills/**` wholesale and 23,164 lines leave the population.

**Ruling: the requirements are right and the manifest is wrong.** The inert allowance is defined as *determines no assertion*. `specInvariants` determine an assertion. The classification follows from the written definition, not from a preference, so there is nothing here to decide on taste.

`agent-skills` is **split, not deleted**:
- skill markdown bodies → inert;
- `skills/**/workflow-assets/**` executable assets → **retained-with-successor**, with a successor ticket and an expiry like any other retained path.

This raises Retained. That is correct. A metric that moves the wrong way because we stopped hiding something is the metric working.

**B4 — the matrix contradicts its headline four times and hides an unclassified file.** `docs/rust-burndown/executable-path-matrix.md` publishes 78,669 at `:88-91`, then says "**81,910** is the burn-down target" at `:104`, shows `23,579 + 81,910 = 105,489 ✓` at `:113-118`, and tells a reader reproducing the numbers to expect "363 files, **105,489** lines" at `:399`. `ed5cceb` refreshed the table and left the prose.

The scripts block header says 13,294 / 38; its rows sum to 13,160 / 37. The gap is exactly `scripts/check-trace-tags.mjs` — 134 lines, added by this branch, **counted in the subtotal with no row written for it**, while `:93` claims "Nothing is unclassified". Re-check the tests and src blocks too; only the scripts block was verified by hand.

`:6` still reads "Measured at: commit `e718d45`" while `:95` calls the figures a fresh measurement "at this revision" and never names it — against this branch's own rule that every figure states its revision.

**B5 — thirteen new requirements, zero matrix rows.** `spec/matrix.md` has no rows for FR-096..FR-103, NFR-024..NFR-027, StR-009, US-024 or IT-003, and no TC rows for TC-1601..TC-1709. The port PRs unblock because the criteria exist, then cannot record coverage because there is no row to mark. **This is why the PR is not split to unblock the ports early** — without the matrix rows the split does not actually unblock anything.

### Important

- **FR-100 requires porting a subsystem this branch deleted.** `7c8e18f` disposed of `src/measurement/{document-state,partition,rates}.ts` and eight siblings. `FR-100:73` requires preserving "measurement states, failure partitions, and published rates"; `FR-100-AC-4` compares against "the retained report", which no longer has a generator. **AC-4 can neither fail nor be satisfied — vacuous by construction, on the branch whose thesis is that vacuous criteria are the enemy.**
- **Dangling references to withdrawn requirements.** `FR-103:12`/`:102` extend and cite FR-084 (⛔ Withdrawn). `FR-100:130` cites FR-090 (withdrawn) and NFR-021/022/023, which remain `🚧 Pending (#291)` — #291 closed, every FR they constrain withdrawn, the code disposed.
- **Two Open dispositions rest on grep claims that are false.** FND-174's "`format` appears nowhere in FR-098" is contradicted by `FR-098:77-79` and AC-8; FND-177's "`strict` appears nowhere" by `FR-098:80-82`. The *decision* to defer both to #378 stands — the stated *reason* does not, and `base.md:81` books both Resolved. Two documents on one branch disagree. (The `allErrors` half of FND-177 is genuinely unaddressed.)
- **FND-186 credits `ed5cceb`**, which re-measured FP-NRX. The 109-test-file staleness it names was fixed by `5968de3`. `base.md:97` has it right.
- **FND-173 and FND-180 are booked as owner rulings in `base.md`** and promoted to Resolved in `dispositions.md`. FND-173 is B3 above and is now ruled. FND-180 needs a dated owner disposition that does not exist in the tree; recording a deferral as a closure is the one move this document exists to prevent.
- **`NFR-024-AC-1` requires an expiry per retained path.** No retained row carries one.

### Correcting two of my own premises

- The set is **NFR-024..NFR-027**, not ..026. NFR-027 (Rust idioms and gates) is real and one of the best documents on the branch.
- **There is no consistent "six allowances" anywhere**: ADR-0003 says two standing-approved categories, NFR-024 enumerates five, the matrix says "1–6", the YAML declares eight. My six was the matrix's six. Getting the artifacts to agree on one number *is* B2.

### What this set gets right — propagate to the port PRs

1. **Named non-empty population + bypass probe as a matched pair.** Not just `FR-100-AC-8`: `FR-098-AC-3` (planted single-byte canonicalization change), `NFR-024-AC-7` (hand edit forfeits the generated exception), `NFR-027-AC-1/5/6`, `FR-103-AC-5`. **Roughly a third of the criteria carry a planted-violation clause.**
2. **Empty-population-is-inconclusive as its own criterion, not a caveat** — `FR-098-AC-7`, `FR-098-CON-3`, `NFR-024-AC-5`, `NFR-025-AC-5`, and `NFR-027-AC-9` generalising it: every check reports the population it scanned.
3. **The self-written-fixture defect raised one level.** FR-101: a fixture recorded as produced by the Rust implementation cannot satisfy a parity criterion. `FR-101-AC-11` requires producing implementation, revision and request digest on every committed fixture.
4. **Digests treated as identifiers, not checksums.** NFR-025: the citing records "do not become wrong in a way a test reports — they become unresolvable." That is why replay is a pre-cutover gate rather than a regression check.
5. **`provisional` (NFR-024-AC-10)** distinguishes not-yet-done from done-wrong inside the metric's own vocabulary, so a regression to an epic-level successor reference is visible rather than silently accepted.
6. **Refusal-path parity as a first-class obligation** — FR-098 ports the strict parser's refusal boundary because it *guards* canonicalization rather than following it, with `FR-098-AC-9` requiring refusal before canonicalization runs.
7. **`NFR-027`: a gate that is described and not run is a green result for a check that did not execute.** AC-8 fails a CI change that removes a lane, narrows one, or adds `continue-on-error`, independently of the code it accompanies.
8. **The manifest's own inert test** — "would changing this file's logic, not its data, change a pass/fail result?" — is exactly right. B3 is that `agent-skills` was created to route around it.

### Disposition

Hold. B1 is one line. B4 and B5 are mechanical. B2 and B3 are settled by the ruling above rather than left open: NFR-024 is amended to enumerate the real category set with one spelling, AC-9 gains the second direction so an unadmitted category fails, AC-6's `owner` becomes a per-entry field, and `agent-skills` splits.

Not splitting the PR to unblock the ports early — B5 means the split would not unblock them.

Confirmed sound and not to be changed: `FR-098-AC-10` names the right store (`src/evidence/assurance-records.ts` produces `sha256:<hex>` ids and `sha256-<hex>.json` names; `src/change-assurance/store.ts` is blake3 over JCS). The `b36156d` revert is clean. `scripts/check-trace-tags.mjs` is deliberately unwired pending the governed relock in #350 — though per B4 it is counted while unclassified.

Issue #384, opened 2026-09-12
[Quoin#384] [Stage 5] quoin-assurance: assurance case construction and discharge


kreneskyp:
Port `src/assurance/` (1,723 lines, 5 files) to a Rust crate `rust/quoin-assurance`.

**The cheapest crate in Stage 5, and the best first real port after `quoin-validators`.** Measured: it has **zero value-level outbound edges** — all five of its cross-directory imports are `import type` (`graph.ts:13,14,15,16`, `discharge.ts:11`). It is in no dependency cycle. Its only inbound edges are from `src/commands/`. It is a pure view module over in-memory inputs, which its own header states (`src/assurance/graph.ts:8`: *"This is strictly a view"*).

Practically: no subprocess, no network, no filesystem, no cycle to wait on. It can be written the day the workspace exists.

## One shared decision, recorded in both leaf tickets

This crate needs exactly four types from auditor — `AuditReport`, `Finding`, `UnevaluatedCheck`, `Severity` (`src/auditor/audit.ts:38-144`) — and nothing else. Three other places want the same four: `src/assurance/graph.ts:14`, `src/graph-analysis/analysis.ts:3`, `src/graph-analysis/input.ts:5`, `src/change-assurance/types.ts:1`.

Once #376 lands, depending directly on `quoin-auditor` is *legal* — no cycle. It is just heavy: a leaf view crate would pull a 1,800-line engine crate to name four structs. **Recommend a small `quoin-finding-types` crate** holding those four. Cheap, and it keeps the view crates genuinely leaf. Not a blocker either way — decide it once, in whichever of these two tickets is picked up first, and make the other follow.

## Generic vs quoin-specific

Generic, and worth asking whether `engineering-assurance` should own it: the GSN-shaped case construction at `src/assurance/graph.ts:109 buildCase` with `NodeKind = "goal"|"strategy"|"solution"` (`graph.ts:22`) is standard assurance-case vocabulary, not quoin's. Same for the deterministic renderers `render.ts:18 renderCase`, `discharge.ts:175 renderDischargeReport`, `argument.ts:512 renderAuthoredArgument` — markdown/JSON projections with a byte-identical-output contract (`render.ts:5`).

Quoin-specific, stays: `discharge.ts:88 buildDischargeReport` and `argument.ts:289 parseAssuranceArgument`, both bound to quire `ClauseSetKey`/`ClauseForce` semantics (`discharge.ts:11`).

[...]

PR #449, opened 2026-09-13
feat(445,447): completeness and assurance across the boundary, and 4,060 lines of TypeScript retired

### LINKED ISSUE #445: [Quoin#445] Stage 3 cutover: wire quoin-completeness through quoin-core dispatch and delete src/completeness/
Parent: #373. Successor to #378. **Gated on #412.**

## Why this exists

#378 delivered `quoin-completeness` as a library crate: 1,447 lines of Rust, a
golden-parity suite captured from the TypeScript oracle, and **zero reachability
from the `quoin-core` binary**. `src/completeness/` — 685 lines — is still the
only implementation anything executes.

That is the program's one standing defect. Across fifteen merged crates and
31,884 lines of Rust, 86.3% cannot be reached from the boundary binary and no
TypeScript has ever been deleted by porting. A library that nothing calls is not
a port; it is a second implementation, and NFR-024 bounds coexistence precisely
because two implementations of one rule is the condition this program exists to
end.

So #378 does not close when PR #402 merged. This ticket is what closing it means.

## The constraint this domain runs into

`quoin-core/tests/tc_library_containment.rs` audits every `quoin-core/src/**/*.rs`
except `main.rs` as a `ReusableLibrary`: the library half must be decidable
**without a filesystem**. `quoin-completeness`'s entry points are
`assess_bundle(bundle_root: &Path, module_roots: &[PathBuf])` and
`load_vocabulary_coverage(&[PathBuf])` — both walk the disk.

An operation that takes a path and reads it is therefore not available, and the
gate is not to be weakened, allow-listed or skipped. The resolution is the shape
`assurance.build_case` already uses: **the request carries content**. The
directory walk and the file reads stay in the command shell — which is where a
CLI's I/O belongs — and every decision crosses the boundary as bytes.

## Scope

1. Give `quoin-completeness` a **content-in** API beside its filesystem one:
   declarations resolved from manifest and schema *text*, frontmatter parsed from
   document *text*, and an assessment computed over those. The existing
   path-taking functions become thin shells over it, so the golden-parity suite
   keeps testing the same code.
2. Register `completeness.*` as command-shaped operations in `quoin-core`
   dispatch — never a bare function. `OPERATIONS` and the match arm both, which
   the #444 drift guard checks against `dispatch.rs`'s own source.
3. Every request type `#[serde(deny_unknown_fields)]`, every accumulator bounded
   in BYTES before any work, refusal named. stdout = payload, stderr =
   diagnostics, exit taxonomy 0/1/2/3/4 mapped from `CoreErrorCode` and nowhere
   else.
4. `schemars` registers the new root types in `quoin-schemas`; `src/core/types.ts`
   is regenerated by `make types` and stays hash-asserted. No hand-written
   response type.
5. Switch `src/commands/completeness.ts` and the `readBundleFrontmatter` caller in
   `src/commands/assurance.ts` over to `runCoreAllowFailure`, **reversibly, as its
   own commit**, demonstrated revertible with `git revert --no-commit`.
6. **Then** delete `src/completeness/` and `tests/completeness.test.ts` — code and
   tests in the same commit, per FR-101.
7. User-visible `quoin completeness` output does not change. Verified by running
   the command before and after on this repository's own `spec/`.

## Acceptance

- Every criterion a deleted TypeScript test carried is restated on a
  tracking-tagged Rust test (`/// Trace:` + `tc_NNN`). A criterion is not carried
  forward by a test asserting against a fixture the test itself wrote, nor by a
  check passing over an empty population.
- No non-Rust runtime oracle survives the cutover. The captured goldens stay; a
  live Node process in the Rust test path does not.
- `make rust-gate` passes.
- The PR states, in numbers, how many lines of TypeScript it deleted.

## Gate

#412 (Stage 2, `quoin-validators`) is the boundary-shape precedent and the epic
states Stage 3 does not start its own cutover until it closes. This lands after
#412, rebased onto whatever boundary shape #412 settles on. One shape across the
program, not two.

### LINKED ISSUE #447: [Quoin#447] Stage 5 cutover: finish the quoin-assurance port, wire it through quoin-core, and delete src/assurance/
Parent: #373. Successor to #384. **Gated on #412.** Lands with #445 in one PR.

## Why this exists

#384 delivered `quoin-assurance` as a library crate. Four of its functions are
reachable from `quoin-core` — `assurance.requirement_of`, `build_case`,
`render_case`, `parse_argument` — and they are the only four operations besides
`core.ping` that the binary answers. Nothing calls them. `src/assurance/` — 1,752
lines — is still the only implementation `quoin assurance` and `quoin discharge`
execute, and `src/core/reference.ts` delegates to it as the difftest oracle.

## What #384 did not port

Measured against `src/assurance/index.ts`'s exported surface, the crate is
roughly half the module:

| retained TypeScript | lines | Rust |
|---|---:|---|
| `graph.ts` `buildCase` / `requirementOf` | 310 | `case.rs`, `lib.rs` |
| `render.ts` `renderCase` | 186 | `render.rs` |
| `argument.ts` `parseAssuranceArgument` | ~160 | `argument.rs` |
| `argument.ts` `buildAuthoredArgumentView` | ~158 | **none** |
| `argument.ts` `renderAuthoredArgument` | ~303 | **none** |
| `discharge.ts` (whole file) | 397 | **none** |

`src/assurance/` cannot be deleted while three of its six exported capabilities
have no successor, so this ticket carries the remaining port as well as the
cutover. It also needs `ClauseSetKey`, `ClauseForce`, `ClauseBinding`,
`ClauseBindingReason` and `ClauseBindingReport` in `quoin-quire-types`, which
holds only `Obligation` today.

## Scope

1. Port the three unported capabilities into `quoin-assurance`:
   `authored::build_authored_argument_view`, `authored::render_authored_argument`,
   and a `discharge` module carrying `build_discharge_report` and
   `render_discharge_report`. The clause-binding types move to
   `quoin-quire-types`, where `src/quire/types.ts`'s successors belong.
2. Register the remaining `assurance.*` operations in `quoin-core` dispatch —
   command-shaped, never a bare function. `OPERATIONS` and the match arm both,
   which the #444 drift guard checks against `dispatch.rs`'s own source.
3. Every request type `#[serde(deny_unknown_fields)]` where a derive can express
   the contract and a hand-written validator where it cannot (the authored
   argument's asymmetric `null` and instant grammar are the precedent from #441);
   every request bounded in BYTES before any work. Exit taxonomy 0/1/2/3/4 mapped
   from `CoreErrorCode` and nowhere else.
4. `schemars` registers the new root types in `quoin-schemas`; `src/core/types.ts`
   is regenerated by `make types` and stays hash-asserted.
5. Switch `src/commands/assurance.ts` and `src/commands/discharge.ts` over to
   `runCoreAllowFailure`, **reversibly, as its own commit**, demonstrated
   revertible with `git revert --no-commit`.
6. **Then** delete `src/assurance/` and its tests — `tests/assurance.test.ts`,
   `tests/authored-argument.test.ts`, `tests/discharge.test.ts`,
   `tests/independence.test.ts`, `tests/trust-decision.test.ts` — code and tests
   in the same commit, per FR-101. `tests/assurance.test.ts` and
   `tests/discharge.test.ts` also cover `src/quire/`, which is NOT in scope; that
   coverage is split out and kept.
7. **Retire the assurance difftest arms to frozen goldens in the same commit.**
   `src/core/reference.ts` is the difftest's TypeScript oracle and per FR-101 it
   *delegates* to `src/assurance/index.js` rather than reimplementing it. Deleting
   the module removes the oracle's subject, and #373 AC-5 forbids a non-Rust
   runtime oracle after cutover. So the retained implementation is run once, its
   answers are captured, and the difftest compares `quoin-core` against those
   bytes thereafter. `core.ping` keeps its live reference — it has no retained
   capability and never did.

## Acceptance

- Old and new pass at **one candidate revision**, in CI, from one checkout, for
  every operation, including the three newly ported ones.
- Every criterion a deleted TypeScript test carried is restated on a
  tracking-tagged Rust test (`/// Trace:` + `tc_NNN`). `quoin-assurance` has
  **zero** lines in its `tests/` directory today, so this is the part of the
  program that needs the closest look: a criterion is not carried forward by a
  test asserting against a fixture the test itself wrote, nor by a check passing
  over an empty population. That failure has shipped four times here already.
- No non-Rust runtime oracle survives the cutover for a retired capability.
- `make rust-gate` passes.
- The PR states, in numbers, how many lines of TypeScript it deleted.

## Gate

#412 is the boundary-shape precedent; this lands after it, rebased onto whatever
shape it settles on. It ships in one PR with #445 because the two interlock:
`src/assurance/graph.ts` imports `BundleDocument` from `src/completeness/`, and
`src/commands/assurance.ts` imports from both.

kreneskyp:
Stage 3 and Stage 5 of the Rust burn-down, in one PR because they interlock: `src/assurance/graph.ts` imported `src/completeness/`, so neither module could be retired without the other.

**4,060 lines of TypeScript deleted.** Two modules and three test files gone outright; the oracle that served them cut from 643 lines to 294.

Was gated on #412 / PR #448 (merged, `8bab74a`), and has since been rebased onto merged `main` at `cf242f1`, which carried #450's `config` and `modules` domains. It adopts #448's boundary shape rather than inventing a second one.

---

## The numbers

| | lines |
| --- | ---: |
| `src/assurance/` (argument 816, discharge 397, graph 310, render 186, index 43) | **1,752** |
| `src/completeness/` (assess 238, bundle 161, declarations 153, run 96, index 37) | **685** |
| `tests/completeness.test.ts` | 368 |
| `tests/assurance.test.ts` | 356 |
| `tests/authored-argument.test.ts` | 267 |
| `tests/discharge.test.ts` (split, not deleted — see below) | −162 |
| `src/core/reference.ts` (643 → 294) | −349 net |
| other | 121 |
| **total TypeScript deleted** | **4,060** |

2,217 lines of TypeScript were added, of which **1,201 are `src/core/types.ts`, which is generated** by `make types` from `quoin-schemas` and hash-asserted. Hand-written TypeScript added: 1,016. Net hand-written TypeScript: **−3,030**.

Ten operations join the boundary:

```
completeness.schema_refs             assurance.build_discharge
completeness.assess_bundle           assurance.render_discharge
assurance.requirement_of             assurance.build_authored_argument
assurance.build_case                 assurance.render_authored_argument
assurance.render_case                assurance.parse_argument
```

## Two commits, never merged

**Commit 1** wires dispatch and rewires the TypeScript callers. It deletes nothing, and reverts cleanly on its own — verified with `git revert --no-commit`, confirmed across all 58 files, undone.

**Commit 2** deletes the TypeScript and its tests together.

## The boundary shape

Adopted from #448: **the request carries content, and the domain crate keeps a source trait.** No operation takes a path and reads the disk, so the library half of `quoin-core` stays decidable without a filesystem and `tc_library_containment` passes unchanged, unweakened and un-allow-listed. The walk and the reads stay in the command shell, reusing `src/core/snapshot.ts` rather than writing a second snapshot helper.

### The pre-filter defect, and what replaced hoping

A TypeScript-side pre-filter that decides what crosses the boundary is verdict-changing for these two domains, because both are analysis over a document tree. Re-classifying on the Rust side protects only against a filter that is too **coarse** — never against one that is too **narrow**, and a test that enumerates the filter's rules on the TypeScript side can only restate what the filter already believes.

So: **one tree, both paths, identical results.**

- `tc_445_320_the_disk_reader_and_the_snapshot_reader_agree_on_one_real_tree` materialises a real tree and asserts `read_bundle_frontmatter(root) == frontmatter_from_sources(independent_snapshot(root))`. The independent walk is written in the test file, not imported from the crate.
- `tc_445_321_the_tree_under_test_is_actually_populated` is the population floor: ≥7 markdown files, ≥6 parseable, ≥1 unreadable, the depth-3 document present by name. A dead probe fails instead of sweeping nothing.
- `tests/core-bundle-snapshot.test.ts` asserts the same equivalence from the TypeScript side, with its own floor, and once **over the real binary**.

Proved by breaking the filter three ways and confirming every probe fails:

| breakage | result |
| --- | --- |
| narrow `bundleSnapshot` by a directory name | 3/3 TS tests fail |
| narrow it by a filename shape (`!e.includes(" ")`) | 3/3 TS tests fail |
| narrow the Rust walker | 2/2 Rust tests fail |

Reverts verified byte-exact by SHA-256.

**It found a real divergence.** A *directory* named `notes.md`: the shell's walk filtered on the extension alone, opened it, and recorded `EISDIR` as an unreadable document the bundle does not contain — an invented document — while `quoin-completeness`' disk reader matches on the file type first and reports nothing. Fixed in `src/core/snapshot.ts` with `withFileTypes` + `isFile()`. A snapshot that adds an entry the tree does not hold is the same class of defect as one that drops an entry it does, and both directions are now pinned.

## The size bound bites at the read

A bound checked after `read_to_string` + a full parse + a `to_vec` has already paid for the memory it exists to refuse. `MAX_REQUEST_BYTES` is now applied through `Read::take` **before** the parse and before any copy; an oversized request costs one byte more than the limit rather than whatever the caller chose to send.

It sits **strictly above** every per-operation bound, and the inequality is load-bearing rather than decorative. Set equal to them, a request one byte past a domain's ceiling tripped the *transport* first and the caller was told `{limit_bytes, op, stream}` where the domain would have said `{limit_bytes, observed_bytes, op}` — the transport pre-empting a refusal that was the domain's to make, and losing the observed size on the way. `quoin-difftest` caught exactly that.

- `tc_445_100` — a request past the ceiling is refused before it is parsed.
- `tc_445_101` — a request **at** the ceiling reaches the operation, proved by the refusal coming from the op's own bound.
- `tc_445_102` — the inequality, over all six operation bounds. Verified it bites: setting the ceiling back to 16 MiB fails with *"assurance.build\_case bounds its input at 16777216, which the transport ceiling 16777216 does not sit above"*.
- `tc_445_103` — a request past an operation bound gets the **operation's** refusal, with `observed_bytes` and no `stream` key.

Tested **at** the size, not only asserted as a constant, because a new ceiling is a new hard failure mode for large repositories.

## `src/core/reference.ts` — the sequencing

`reference.ts` is the TypeScript oracle `quoin-difftest` compares against, and per FR-101 it *delegates* to the retained `src/assurance/index.js` rather than reimplementing it. Deleting `src/assurance/` removes the oracle's subject, and FR-101-AC-5 forbids a non-Rust runtime oracle after cutover. The honest sequence, following the precedent #448 set:

1. **Commit 1 ADDS four assurance oracle arms** (`build_discharge`, `render_discharge`, `build_authored_argument`, `render_authored_argument`), so the difftest proves the new operations at parity **while their subject still exists**: 123 cases → 168, zero differences.
2. **Commit 2 REMOVES all eight**, netting zero against main.
3. The criteria are replayed as a **frozen golden corpus** through the real binary.

`core.ping` keeps its live two-implementation reference, because it never had a retained TypeScript capability to lose.

`quoin-difftest` therefore drops from 168 cases to 12. **That is coverage that moved, not coverage lost:** `rust/crates/quoin-assurance/tests/golden/` holds 128 hand-authored cases whose verdicts were captured once from the retained TypeScript, replayed against frozen bytes by `tc_447_500..502` in the library and through `quoin-core` over a pipe by `tc_447_510..512`, each with an anti-vacuity length floor. `PROVENANCE.md` states the oracle revision, the six source SHA-256s, the population in numbers, and three divergences.

The capture's generator is deleted with its subject. A script importing a module no longer in the tree is a dormant Node oracle with a longer fuse; `PROVENANCE.md` records the pre-deletion revision and the hashes that let a reader re-derive the bytes there instead.

### The corpus population, stated

Because a comparison that runs over half a population is green and says nothing. 128 cases: `requirement_of` 14, `build_case` 33, `render_case` 29, `parse_argument` 52. 81 captured a value, 47 captured a throw. Of the 32 `build_case` cases that returned, 5 supply a non-empty `findings` list and 27 supply none; 13 produce a case with at least one `open` node and 7 are wholly `supported`; 12 produce an empty case and 20 a populated one. 13 cases are inputs the Rust request schema refuses. **`cases.json` is hand-authored and every expected value came out of the TypeScript** — this is not a fixture the Rust tests wrote for themselves, and two assertions failed on first run.

### Three TypeScript↔Rust divergences, recorded not papered over

1. **The request schema refuses input the retained library accepts (8 cases).** Unknown request fields, a non-string `obligation_id`, a `CaseNode.kind` outside the closed set, a `Finding` missing `kind`. Plain JavaScript ignores or coerces; the port types all four, so serde refuses with `CORE_BAD_REQUEST`. This is the port declining to reproduce a coercion. Their TypeScript verdicts are still captured so the difference stays in the bytes.
2. **`build_case` carries assessments `render_case` will not read (2 cases).** `producer_trust` / `evidence_independence` are opaque to `build_case` on both sides, so a case can carry an assessment missing `useId` or whose `triggeredBy` holds objects. The retained renderer interpolates `undefined` and `[object Object]`; the port's typed `TrustAssessment` refuses. 27 of 29 built cases re-render at the boundary; 2 do not. `tc_447_512` asserts both halves.
3. **A non-object request is refused one layer earlier** — `CORE_BAD_JSON` from the dispatcher rather than `CORE_BAD_REQUEST` from the handler. Same exit status, different code, contractual.

## Criteria restatement before deletion

No vitest file was deleted until every criterion it carried was restated on a tracking-tagged Rust test (`/// Trace:` + `tc_NNN`). `quoin-assurance` had **zero** lines in `tests/` when this started, so it got the closest look:

- `tests/case.rs` — 18 tests, `tc_447_400`..`417`, covering FR-040-AC-1..15, FR-093-AC-7, FR-094-AC-7/8. Strengthened beyond the TypeScript: whole-`CaseNode` equality, the exact `because` sentence, whole rendered documents.
- `tc_447_412`/`413` found that the retained AC-15 fixture **never actually straddles the UTF-16 boundary**. The half Rust can get wrong is now asserted: `node_id` emits two underscores per astral character, pinned as the exact line `  FR_001_AC___(["FR-001-AC-😀: It holds."])`, with the `chars()`-port spelling asserted absent.
- `tests/discharge.rs` — FR-046.
- `quoin-assurance/src/lib.rs` — the five `requirement_of` unit tests renamed and tagged `tc_447_420`..`424`.
- `quoin-completeness/src/declarations.rs` — `tc_445_230`/`231` close two FR-037-AC-2 branches the TypeScript never reached.

**FR-101-AC-4 wants criteria stated over the shipped command**, so `tests/completeness-cli.test.ts`, `tests/independence.test.ts` and `tests/trust-decision.test.ts` drive the real oclif commands against the real binary, and the assurance corpus is replayed through `quoin-core` over a pipe rather than only in-process.

### Two tests rescued rather than deleted

- **`tests/discharge.test.ts`**: 238 → 87 lines. Its `parseClauseBinding` test is the repo's only coverage of that function in the **retained** `src/quire/validate.ts`, which is not in this PR's scope. Split, not deleted.
- **FR-037-AC-13** in `tests/completeness.test.ts` calls `assessBundle`, which survives in `src/core/completeness.ts`. Deleting the file would have left the criterion asserted nowhere — and it is specifically the criterion about two implementations not drifting. Lifted into `tests/completeness-cli.test.ts`, where it now runs over the boundary.

## User-visible output

`quoin completeness` and `quoin assurance` human-readable output is **byte-identical** to the pre-cutover baselines (`diff` exit 0 both, the assurance document being 724 KB of it).

**`--json` key order changes, and I am saying so rather than claiming output is unchanged.** Canonical JSON sorts keys, so a payload re-emitted from the boundary's stdout comes out alphabetical:

| | before | after |
| --- | --- | --- |
| top level | `bundleRoot, vocabularies, unresolved, unreadable, rollups, findings, verdict` | `bundleRoot, findings, rollups, unreadable, unresolved, verdict, vocabularies` |
| `rollups[]` | `vocabulary, declared, owned, excused, unowned` | `declared, excused, owned, unowned, vocabulary` |
| `findings[]` | `vocabulary, value, kind, severity, message` | `kind, message, severity, value, vocabulary` |

**The same reorder reaches `quoin assurance` and `quoin discharge`, and the table above originally named only `completeness`.** `src/commands/assurance.ts:140,217` and `src/commands/discharge.ts:64-65` re-emit the boundary payload too, so every node the argument view prints is alphabetised:

| | before | after |
| --- | --- | --- |
| argument node (`src/assurance/graph.ts:267` on `main`) | `id, kind, statement, status, because?, children` | `because, children, id, kind, statement, status` |

Same conclusion as above: values and field names are unchanged, so a consumer that parses is unaffected and a consumer that diffs raw bytes will see it. Restoring the old order would mean hand-writing a field list, which FR-097 forbids.

`assert.deepStrictEqual` passes after parse: the values are identical and only the serialisation order moved. Any consumer parsing the JSON is unaffected; a consumer diffing the raw bytes will see it.

## Generated type names

`FindingKind2` in the generated `src/core/types.ts` was a `schemars` disambiguation artefact from the collision between `quoin_completeness::FindingKind` and `quoin_validators::FindingKind`. A reader could not tell which domain's finding kind it was, and **which type got the `2` depended on registration order in `boundary_schema()`**, so adding a third domain could silently rename an exported type. Fixed rather than flagged: the completeness enum now carries `schemars(rename = "CompletenessFindingKind")`, so the generated surface names a domain instead of a collision ordinal, and `the_schema_covers_every_type_that_crosses_the_boundary` pins the whole surface against it.

## Gates

Re-run in full **after the rebase onto `cf242f1`** (merged #450), not before.

| gate | result |
| --- | --- |
| `make rust-gate` | **pass** — fmt, clippy `-D warnings`, cargo deny, tests, difftest |
| `make rust-difftest` | 12 cases, **0 differences** (168/0 at commit 1, before the oracle retired) |
| `make rust-e2e` | **pass** — 9 files, 56 tests |
| `make lint` | **pass** — typecheck, import cycles, eslint, prettier |
| vitest, whole suite | 1088 passed, 9 skipped, **6 failed for want of a checkout** |

The six vitest failures are environmental, not branch-caused: `corpus/` is a submodule that is not initialised in this worktree (3 cases in `bench-tier1.test.ts` read `corpus/config/metrics.json` and `corpus/bounds.py`), and `semantic-module-template.test.ts` needs `spec-objects-business` cloned beside the repository (2 cases). A seventh failure seen before the re-run — `cli.test.ts`'s "the default module set is reachable from the built entry" — was a stale `dist/`; it passes after `make build`.

`make test` repo-wide is still blocked by #430: its `verification-stack.mjs` prelude wants a `quire-rs` checkout beside the tree. vitest was therefore invoked directly.

One clippy `items_after_statements` error in `quoin-schemas/src/render.rs` was fixed by hoisting `ANNOTATIONS` to module scope — not by an `#[allow]`. No pin weakened, no lint relaxed, no gate lowered anywhere in this PR.

## After the rebase onto `cf242f1`

The rebase brought #450's `config` and `modules` domains under this branch's own guards, and three of them fired — which is the point of them:

- **The bound census** compared `ops/mod.rs` against a flat `OPS_SOURCES`, and the two new domains are *directories*. It now names each source file beside its domain and checks each directory's `mod.rs` against the files listed under it, so a bound added in a new `ops/modules/<new>.rs` cannot go uncensused either. `tc_412`'s transport census gained the six byte bounds those domains declare.
- **`tc_447_600`** found four operations dispatched with no test invoking them *by name*: `config.resolve_org`, `config.unresolved_org_message`, `modules.list`, `modules.remove`. A swapped match arm would have routed them to another handler with every gate green. `rust/crates/quoin-core/tests/tc_449_config_modules_boundary.rs` is that route coverage; the remove case checks the registry rather than the payload that claims the removal.
- **The boundary type surface** still listed `FindingKind2`, resolved as above.

Also restored the line continuation `rust-e2e`'s file list lost in the rebase, which made `make` run the second half of the list as a command.

Closes #445
Closes #447
Refs #373, #378, #384

## Review cycle (independent review, 2026-09-12)

Five findings, all addressed in this PR:

1. **HIGH — eight FR-037 criteria lost their only backing.** `tests/completeness.test.ts` carried `FR-037-AC-1`, `CON-1` and `AC-3..AC-8`; the Rust tests covering the same behaviour carried a bare `FR-037`, which binds the requirement and no criterion, so `quire coverage` reported TC-206 and TC-208..TC-213 as `status_lies` — a ✅ over nothing. Seven were a tagging fix (`tc_378_301` already grids kind AND severity with a `kinds_seen` floor; `tc_378_210`/`211`; `tc_378_213`). AC-1/CON-1 had no test at all — nothing asserted the vocabulary's VALUES, only that declarations resolved, which a minted list satisfies too — so `tc_449_240` reads two declarations off two different schema enums. Verified: `quire coverage --json` now reports zero unbacked FR-037 criteria and none of those rows in `status_lies`.
2. **MEDIUM — self-referential census.** `tc_412_the_transport_ceiling_stays_above_every_domain_bound` pinned itself with `assert_eq!(domain_bounds.len(), 6)`, where `domain_bounds` is the array literal directly above it: the quoin#443 shape. A bound added to an `ops` module and never listed left the inequality unevaluated and the test green. The census is now compared against the `ops` sources via `include_str!` (the #444 drift-guard shape, no host capability), with `ops/mod.rs` as the independent authority on which modules exist and a >= 4 anti-vacuity floor. Both halves were broken and observed to fail.
3. **MEDIUM — undisclosed `--json` key-order change on the assurance commands.** Disclosed above.
4. **LOW — stale counts.** The difftest module doc said 156 assurance cases; 111 were retired.
5. **LOW — PR-body gate results predate the rebase onto `cf242f1`.** Re-run and restated after the rebase.

Probed clean by the reviewer, with the breakages performed and restored: the pre-filter superset property on both axes and both sides, the golden replay, the operation census, and the size bounds biting before the parse.


kreneskyp:
## Independent review — #449 (reviewer pass, not the author's)

Scope: the diff against `origin/main` only. Gates re-run read-only in a throwaway worktree: `make rust-gate` exit 0, `make lint` exit 0, full vitest (excl. `bench-tier1`) 99 files / 1095 tests pass, 9 skipped. Every claim below marked CONFIRMED was produced by breaking the thing and watching the named test fail, or by a measurement — not by reading.

### First: what the difftest corpus went from 123 to 12 **by**

`quoin-difftest/src/main.rs` on `origin/main` registers 123 cases: `assurance.build_case` 25, `assurance.parse_argument` 51, `assurance.render_case` 24, `assurance.requirement_of` 11, `core.ping` 11, `evidence.record` 1. This branch registers 12: `core.ping` 11 + `evidence.record` 1. **Exactly 111 cases were removed, all four `assurance.*` arms, and nothing else.** That is correct under FR-101 AC-5 — `src/core/reference.ts` loses its assurance oracle (−412/+63, only `core.ping` retained), so there is no live non-Rust oracle left to differentiate against.

Where the 111 went: I name-mapped all 111 against `quoin-assurance/tests/golden/cases.json` (128 cases). **105 are represented; 6 are not, and all 6 are the exclusions PROVENANCE.md already declares** — 4 transport-level malformed/non-object inputs (not representable as a golden case, covered by `ops::assurance` unit tests) and the 2 16 MiB size-boundary cases (covered by `tc_445_100..103`). Per-op, the golden corpus is strictly larger than what it replaced on every operation: requirement_of 14 vs 11, build_case 33 vs 25, render_case 29 vs 24, parse_argument 52 vs 51.

Population floors exist on both replays and **both fire** — truncating `cases.json` to 10 cases failed `tc_447_500` ("the golden corpus came back with 10 cases; it is not being read") and `tc_447_501` ("the corpus holds 10 requirement_of cases; it was authored with at least 12"). The corpus is replayed through the real binary over a pipe by `quoin-core/tests/tc_447_assurance_boundary.rs::tc_447_510`, not only in-process. This is a sound swap.

---

## Findings

### 1. HIGH — CONFIRMED by measurement. Eight FR-037 criteria lost their only backing.

`tests/completeness.test.ts` (deleted) carried `// Trace: FR-037-AC-1, FR-037-CON-1`, `-AC-3`, `-AC-4`, `-AC-5`, `-AC-6`, `-AC-7`, `-AC-8`. After this PR **nothing in the tree names any of them.** The Rust tests that cover the behaviour (`quoin-completeness/src/assess.rs::tc_378_210..213`, `declarations.rs`, `run.rs`, `bundle.rs`) carry only the bare `/// Trace: FR-037` — a requirement, not a criterion — and `git diff` on those Trace lines is empty, so they are pre-existing #378 tests this PR did not re-tag.

The engine agrees. `quire coverage --scope . --json` on this branch reports as **`status_lies`** (matrix says ✅, nothing backs it):

```
spec/matrix.md:454  TC-206  ✅  FR-037-AC-1
spec/matrix.md:456  TC-208  ✅  FR-037-AC-3
spec/matrix.md:457  TC-209  ✅  FR-037-AC-4
spec/matrix.md:458  TC-210  ✅  FR-037-AC-5
spec/matrix.md:459  TC-211  ✅  FR-037-AC-6
spec/matrix.md:460  TC-212  ✅  FR-037-AC-7
spec/matrix.md:461  TC-213  ✅  FR-037-AC-8
```

plus `FR-037-CON-1` unbacked at `spec/functional/FR-037-declared-vocabulary-completeness.md:118`, and the matching `verification` rows at lines 100–107. FR-037-AC-2 and AC-9..13 are fine (AC-9..13 were correctly lifted into `tests/completeness-cli.test.ts`, which genuinely drives the oclif command). FR-040, FR-046 and FR-047 are all restated at criterion level. **FR-037 is the sole gap, and it is seven matrix rows plus a constraint reading green over nothing.**

This is what #445's acceptance forbids ("Every criterion a deleted TypeScript test carried is restated on a tracking-tagged Rust test") and what the PR body claims was done.

Fix is tagging, not new tests — the behaviour is already covered, e.g. `tc_378_210` covers AC-7 and `tc_378_213` covers AC-8. Extend the `/// Trace:` lines on the existing `quoin-completeness` tests to name the criteria.

### 2. MEDIUM — CONFIRMED. The bounds census pin is self-referential.

`rust/crates/quoin-core/src/dispatch.rs` (~415):

```rust
assert_eq!(
    domain_bounds.len(), 6,
    "a domain bound was added or removed without this census moving"
);
```

`domain_bounds` is the hand-written array literal on the lines directly above, so `len()` is `6` by construction and the assertion cannot fail. Scenario: a new op in `ops/semantic.rs` declares `MAX_SEMANTIC_BYTES = 128 * 1024 * 1024` — above the 64 MiB transport ceiling — and is not added to the array. The census stays green, and the `bound <= MAX_REQUEST_BYTES` inequality it exists to enforce is never evaluated for that bound; the op refuses nothing until the transport layer truncates.

The correct shape is two files away: `ops/completeness.rs::a_third_completeness_error_code_would_need_a_mapping` pins against `CompletenessErrorCode::all().len()`, a value the compiler owns. Derive `domain_bounds` from something the compiler owns, or drop the count and assert only the inequality per entry.

### 3. MEDIUM — CONFIRMED by inspection. Undisclosed `--json` key-order change on two more commands.

`src/commands/assurance.ts:140,217` and `src/commands/discharge.ts:64-65` now re-emit the boundary payload with `JSON.stringify(x, null, 2)`. `quoin-core` emits canonical JSON (`serde_json::Map` is a `BTreeMap`, `preserve_order` deliberately off), so keys are alphabetical at every depth on the way out. On main, `src/assurance/graph.ts:267` builds a node as `{id, kind, statement, status, because?, children}`; it now reaches the user as `{because, children, id, kind, statement, status}`.

The PR body discloses this for `quoin completeness` with a three-row table and says nothing about `quoin assurance` or `quoin discharge`. Values and field names are unchanged (`AssuranceCase` carries `#[serde(rename_all = "camelCase")]`), so a JSON-parsing consumer is unaffected — a byte-diffing or snapshot-asserting one is not. #448 shipped exactly this undisclosed. Add the two commands to the disclosure table.

### 4. LOW — the PR body is stale in three places.

- "fails on 5 files that are pre-existing on `origin/main`" for `make lint` — `make lint` exits **0** on this branch.
- "+13,877/−5,230" — the PR is now **+17,208/−6,350**.
- `quoin-difftest/src/main.rs` module doc says "the 156 assurance cases"; that is the commit-1 peak. Relative to `origin/main` it is **111**.

### 5. LOW — notes, no action required

- `ops/mod.rs::request_size` re-serialises the whole `Value` (up to the 64 MiB transport ceiling) to measure it before comparing against the 16 MiB domain bound. The transport ceiling already caps it, so this is bounded, but it is a second full copy of a payload that is about to be refused.
- `src/core/snapshot.ts::bundleSnapshot` has no directory-exclusion list at all, unlike `repoSnapshot`. A bundle root containing `node_modules` or `.git` would be crossed in full.
- `quoin-schemas/src/render.rs` grew 578 → 987 lines. Below the seam-splitting threshold the rest of the PR applied, but it is now the largest file in the crate.
- FR-037-AC-9..13 are backed only inside `describe.skipIf(coreBinary() === null)`, i.e. they run in the `make rust-e2e` lane and are skipped by a bare `vitest`. That is the established `core-exec-e2e` pattern, not a regression — flagging it because the criteria in question are exactly the "matrix ✅ over a capability no invocation reaches" case the file's own header warns about.

---

## What I probed and found clean

**Pre-filter superset property (both axes, both sides) — CONFIRMED to bite.** Four independent narrowings, each restored:

| Narrowing | Result |
|---|---|
| `quoin-completeness/src/bundle.rs` walker skips dir `level2` | `tc_445_320` FAILED with a full `FrontmatterRead` left/right diff naming the three missing entries; `tc_445_321` FAILED: "the disk reader yielded 5 parseable documents, expected at least 6" |
| same walker drops filenames containing a space | `tc_445_320` FAILED |
| `src/core/snapshot.ts::bundleSnapshot` skips a directory by name | 3 of 4 tests in `tests/core-bundle-snapshot.test.ts` FAILED, including the over-the-real-binary one at :339 |
| same, drops filenames containing a space | 2 of 4 FAILED |

Both the filename-shape axis and the directory-exclusion axis bite, on both the Rust and the TypeScript walk, and the independent walks really are independently written (`symlink_metadata`/`metadata`, `lstatSync`/`statSync`).

**Golden replay bites.** Neutering `quoin_assurance::requirement_of` to `return obligation_id;` failed `tc_447_500` with "requirement-of/ac-suffix diverged from the captured TypeScript verdict / left: `{"requirement": "FR-001-AC-3"}` right: `{"requirement": "FR-001"}`". Narrowing it with `starts_with("fr")` failed `tc_447_500` **and** `tc_447_510` (real binary, over a pipe) identically. Golden SHA-256s re-verified against PROVENANCE.md after restoring.

**Operation census bites.** Repointing the `completeness.schema_refs` dispatch arm at `read_frontmatter` failed `tc_447_604` with the boundary's own diagnostic: `CORE_BAD_REQUEST … unknown field 'manifest', expected 'documents' or 'unreadable'`. `tc_447_600` derives its op list from `OPERATIONS` itself with anti-vacuity floors (`sources.len() >= 4`, `OPERATIONS.len() >= 13`) — a sound census.

**Size bounds bite before the allocation.** `dispatch::read_request` reads bytes into a `take(ceiling)`-capped buffer, compares the byte count, and only then decodes UTF-8 and parses — the #448 defect (`read_to_string` + parse + `to_vec` before the check) is not present here.

**`main.rs` is untouched** — zero diff, no policy added to the containment escape hatch.

**No gate weakened.** No workflow file touched; no lint relaxed, no pin loosened, no threshold lowered; no `push`/`pull_request` trigger added. The two new `#[expect(clippy::too_many_lines, reason = "...")]` sites are precedented on main and carry reasons.

**Exit taxonomy.** Assurance's two error classes are one-field tuple structs, not enums, so there is no `_`-arm-over-`#[non_exhaustive]` hazard; `tc_447_324` obtains each error by type (not by message) and asserts both the `CoreErrorCode` and the `Outcome`. Completeness pins against `CompletenessErrorCode::all().len()`. Clean.

**Deletions.** Every deleted test file read via `git show origin/main:<path>`, criterion by criterion. FR-040 (AC-1..15, CON-2), FR-046 (AC-1..6), FR-047 (AC-1..7) are all restated on tracking-tagged Rust tests. `tests/discharge.test.ts` was correctly **split, not deleted** — it retains FR-046-AC-1 over `parseClauseBinding`, which still lives in `src/quire/validate.ts`. No deleted file covered a module that still exists; the deleted files only type-imported from `src/quire`/`src/auditor`. FR-037 (finding 1) is the single exception.

**Hygiene / the split.** Real and along domain seams, not an arbitrary chop: `argument/{types 259, instant 282, read 154, parse 458, mod 56}`, `authored/{view 361, build 610, render 286, parse 440, fixtures 116, mod 51}`, `discharge/{report 461, build 326, render 84, parse 353, fixtures 63, mod 91}`, tests into `case_graph.rs` 365 / `case_rendering.rs` 219 / `case_assessment_context.rs` 188 over a shared `common/`. Each module's `//!` header names one concept it owns. Nothing in the touched set exceeds 1000 lines.

Finding 1 is the only thing I would hold the merge on, and it is a tagging change.

PR #450, opened 2026-09-13
feat(446)!: config and modules across the boundary, and 1,267 lines of TypeScript retired

### LINKED ISSUE #446: Stage 7 cutover: wire quoin-config and quoin-modules through quoin-core dispatch and retire the TypeScript path
Parent: #373. Successor to #381. **Must land before Stage 8 begins.**

## Why this exists

#381 delivered `quoin-config` (2,697 lines) and `quoin-modules` (2,974 lines) as
library crates and closed. Neither is reachable from the `quoin-core` binary: no
dispatch arm, no operation name, no TypeScript caller, no generated types, no
deleted TypeScript. PR #399 merged 5,671 lines of Rust and removed nothing.

That is the same decomposition #412 exists to close for `quoin-validators`, and
for the same reason: a crate that nothing calls is not a port, it is a second
implementation. The EPIC's own measure (FR-101) is that old and new are exercised
at one candidate revision and then the old one is **deleted**. Across fifteen
merged crates and 31,884 lines, zero lines of TypeScript have been deleted by
porting. This ticket is what closing #381 means.

## Targets (measured)

| Rust crate | crate src | TypeScript it replaces | lines |
|---|---:|---|---:|
| `quoin-config` | 2,697 | `src/config-schema.ts` | 56 |
| | | `src/org.ts` | 261 |
| `quoin-modules` | 2,974 | `src/modules.ts` | 36 |
| | | `src/plugins.ts` | 248 |

`src/modules.ts` and `src/plugins.ts` import each other, so they move together.

`src/catalog.ts` (212 lines) is **out of scope and stays**. `quoin-config`'s
`paths.rs` and `quoin-modules`' `paths.rs` both port its `ixHome` /
`filamentModulesDir`, but `catalog.ts` also owns `loadCatalog`, which
`src/write.ts` and the semantic commands still call. It is not a leaf, and
half-deleting it is worse than leaving it to its own ticket.

## Scope

1. Register `config.*` and `modules.*` as **command-shaped** operations in
   `quoin-core` dispatch — `<domain>.<op>`, never a bare function. Exposing
   `originOrg` or `parseSourceArg` over IPC creates a boundary that can never be
   deleted; exposing `config.resolve_org` and `modules.install` does not.
2. stdout = payload, stderr = diagnostics. No diagnostic on stdout, ever.
3. Full exit taxonomy 0/1/2/3/4. Map `ConfigErrorCode` and `ModulesErrorCode`
   onto `CoreErrorCode` deliberately and **test the mapping**. Both upstream
   enums must be guarded by a pinned `ALL.len()` count, not by a loop over `ALL`
   that re-runs the same match and agrees with itself (quoin#443's failure mode).
4. Register the new root types in `quoin-schemas`; `src/core/types.ts` is
   regenerated and hash-asserted, never hand-edited.
5. Resolve the `tc_library_containment` constraint **honestly**. Config and
   modules are I/O-heavy by nature — config files, a plugin registry, git — and
   the library half of `quoin-core` must stay decidable without a filesystem.
   Do not weaken, skip, allow-list or relax that test. State the resolution in
   the PR. Reuse #412's resolution rather than inventing a second boundary shape.
6. Switch the TypeScript callers over **reversibly, as its own commit**,
   demonstrated revertible with `git revert --no-commit`.
7. **Then** delete the retained TypeScript and its tests — code and tests in the
   same commit, per FR-101.

## Acceptance

- Every criterion a deleted TypeScript test carried is restated on a
  tracking-tagged Rust test (`/// Trace:` + `tc_NNN`). A criterion is not carried
  forward by a test asserting against a fixture the test itself wrote, nor by a
  check passing over an empty population. This program has shipped that failure
  four times.
- User-visible command output is unchanged — verified by running
  `quoin module list`, `quoin module install`, `quoin module remove`,
  `quoin catalog list` and `quoin write` before and after.
- The PR states, in numbers, how many lines of TypeScript it deleted.

## Gate

Stage 8 does not begin until this ticket closes. As with #412, merging the
crates unwired was acceptable; the cutover is what waits.

kreneskyp:
Stage 7 of the Rust burn-down (#373), the cutover ticket #446: `quoin-config`
and `quoin-modules` are wired through the `quoin-core` boundary and the
TypeScript they replace is deleted.

**1,267 lines of TypeScript deleted, across eight files:**

| File                             | Lines |
| -------------------------------- | ----: |
| `src/org.ts`                      |   261 |
| `src/plugins.ts`                  |   248 |
| `src/modules.ts`                  |    36 |
| `tests/org.test.ts`               |   390 |
| `tests/plugins.test.ts`           |   166 |
| `tests/org-no-subprocess.test.ts` |    59 |
| `tests/props/fr-025.prop.test.ts` |    69 |
| `tests/props/fr-018.prop.test.ts` |    38 |
| **Total**                         | **1,267** |

plus the FR-018-AC-1..AC-4 and FR-025-AC-2/AC-3/AC-11 sections of
`tests/props/second-pass.prop.test.ts`, which were grounded on those modules.

## Four commits, never merged (FR-101)

1. `feat(core)` — the dispatch arms and the rewired callers. Verified
   reversible on its own, again after the latest rebase: `git revert
   --no-commit` of commit 1 at commit 1 lands on exactly the base tree
   (`git diff --cached --stat 8bab74a` is empty).
2. `feat(core)!` — the TypeScript **and its tests**, deleted together.
3. `fix(core)` — the six `/rust-review` findings.
4. `fix(core)` — the nine independent-review findings.

## The library-containment constraint, resolved honestly

`rust/crates/quoin-core/tests/tc_library_containment.rs` is **untouched**. No
allow-list, no skip, no relaxation. The library half of `quoin-core` still names
no filesystem, no child program and no network.

Following the shape #448 settled on: **the request carries content, and the
domain crate keeps a source trait.**

- `config.resolve_org` is handed the three configuration documents' *content*
  in the request. It opens nothing. `quoin-config` keeps
  `resolve_org_from_documents` beside the disk-reading `resolve_org`, and
  `tc_446_org_parity.rs` proves the two agree.
- A module install genuinely cannot be expressed as content-in: it fetches a
  git remote, materialises a subtree and rewrites a registry. **Saying so
  explicitly, as asked:** those bytes do not ride on stdin. `ops::modules` is
  therefore *granted* a `ModuleHost` capability object built in `main.rs` —
  the one file the containment audit exempts — and decides everything it can
  decide without one (source-argument parsing, manifest validation, error
  mapping, size refusal) with no disk and no network. Every unit test in
  `ops/modules/tests.rs` runs against an in-memory host, so containment holds
  by construction rather than by exemption. See
  `rust/crates/quoin-core/src/capabilities.rs`.

## FR-025-AC-7 was amended, not dropped

The criterion read "Resolution SHALL NOT execute any subprocess." The cutover
made that **false of the shipping CLI** — resolving an org now calls
`quoin-core` once — and a criterion that is false about the code is worse than
no criterion.

It is amended to the property it was written against: no `git` executable, no
program found on `PATH`, exactly one subprocess and it is the boundary, and
that call is handed the Git configuration's *content* rather than a path.
Resolution still works on a host with no Git at all (NFR-004). The preceding
bullet — "SHALL NOT invoke `git`" — is unchanged and still true.

Both halves have a real carrier:

- `rust/crates/quoin-config/tests/tc_446_org_no_child_program.rs` —
  `tc_446_070` audits `src/org.rs` for a child-program or network capability
  with the **same** `engineering_assurance::source_audit` auditor
  `tc_library_containment` uses, at the same workspace pin.
  `tc_446_071` is the control: it proves the audit *fails* when a `git`
  subprocess is named, so `tc_446_070` is not a check over an empty population.
  It also asserts the file still holds the resolver, so an audit pointed at the
  wrong file fails rather than reporting clean.
  The audit is deliberately narrower than `tc_library_containment` (one file,
  two capabilities), and the module doc says why: `org.rs` legitimately names
  `Filesystem` for the in-process `resolve_org`, and `std::process::id()` in
  `lock.rs` / `service.rs` is flagged `ChildProgram` by the auditor, so a
  whole-crate audit would have needed an allow-list — the rot this program
  forbids.
- `tests/org-one-subprocess.test.ts` — successor to the deleted
  `tests/org-no-subprocess.test.ts`. It replaces **every** `node:child_process`
  entry point: the one `src/core/exec.ts` uses with a recorder that still runs
  the real call (so a real resolution happens), and the other six with
  recorders that throw. It asserts one spawn, `quoin-core`, and never `git`.
  Not done by emptying `PATH`: `execFileSync("/usr/bin/git", …)` succeeds with
  an empty `PATH` and would pass such a test unchanged. It runs in
  `make rust-e2e`.

## No criterion was deleted with its test

Every criterion the deleted tests carried is restated on a tracking-tagged Rust
test over a real population — not a fixture the test wrote for itself, not a
check over an empty set:

| Criteria             | Carrier                                                                       |
| -------------------- | ----------------------------------------------------------------------------- |
| FR-018-AC-1..AC-5    | `quoin-modules` `tc_446_060`..`tc_446_065`, `ts_oracle` `tc_381_230`           |
| FR-019-AC-1/2/4      | `quoin-modules` `tc_381_280`, `tc_381_237`                                     |
| FR-023-AC-4/5        | `quoin-config` `tc_381_104`, `tc_446_026`                                      |
| FR-025-AC-1..AC-11   | `quoin-config` `tc_381_100`..`tc_381_107`, `tc_446_026/027`, `tc_446_050`..`tc_446_053` |
| FR-027-AC-1..AC-5, AC-9 | `quoin-config` `tc_381_103`, `tc_446_026/027`, `config_service.rs`          |

28 Verification-column rows that named a deleted file now name the carrier, in
FR-018, FR-019, FR-023, FR-024, FR-025, FR-027, NFR-003 and `spec/matrix.md`.
Every cited carrier was grepped before being written down.

## File structure

Per the owner's hygiene instruction, `ops/modules.rs` (1,002 lines) was split
along the seams its own doc comments already changed subject at, rather than at
a line count:

```
ops/modules/mod.rs        286  the four operations and the helpers they share
ops/modules/wire.rs       181  request/payload shapes and the ceilings they are read under
ops/modules/taxonomy.rs   181  the ONE place a ModulesError becomes an exit status
ops/modules/support.rs    188  the in-memory host both test files run against (cfg(test))
ops/modules/tests.rs      276  behaviour of the four operations (cfg(test))
```

`ops/config.rs` became `ops/config/mod.rs` (527 lines, 282 production) beside a
new `ops/config/taxonomy.rs` (178), the twin of `ops/modules/taxonomy.rs`. The
`mod.rs` content is the old file unchanged apart from the `mod taxonomy;` line;
the split exists so the `ConfigErrorCode` mapping and count pin have the same
home their `ModulesErrorCode` counterpart has.

`ContractGate` moved out of `main.rs` into
`rust/crates/quoin-modules/src/contract_gate.rs` (668 lines, ten unit tests).
`main.rs` is **219 lines**, down from 486: `main`, `run`, `HostModules`, the
`ModuleHost` impl, and the two emitters.

## User-visible output

Commands were run before and after. Four differences, each explained rather
than smuggled:

1. `quoin module install` now prints the semantic pin in its success line —
   the Rust installer records it and the TypeScript did not.
2. The git cache moved from `ts-plugin-kit/git` to
   `quoin-modules/content/<url>/<sha>`.
3. `path:` with an empty remainder is now refused (QM001) instead of being
   accepted as an empty path.
4. Git transport failure text is `gix`'s rather than `simple-git`'s.

Exit codes are unchanged in all four. **Canonical-JSON key order:** checked, as
asked. No command re-emits a payload parsed from the boundary's stdout into its
own `--json` output; the boundary's canonical ordering is consumed into typed
values and re-rendered by the existing TypeScript emitters, so no `--json` key
order changed.

## Scope decisions stated explicitly

- **`src/catalog.ts` is OUT of scope.** It is not a clean leaf — it reaches
  into the plugin registry and the Ink renderers — and half-deleting it would
  be worse than leaving it. It belongs to a later ticket.
- **`src/config-schema.ts` is BLOCKED**, not skipped. Its `ixSchema`/zod
  surface is consumed by `ix-cli-core`'s FR-014 plugin registry and by the Ink
  renderers; retiring it is Stage 9's work, not this ticket's. This is the one
  part of the measured scope not completed, and nothing else was scaled down.
- **FR-027-AC-8** is left true and is inherited by Stage 9.

## Other facts worth stating

- `QM022_SEMANTIC_CONTRACT_UNAVAILABLE` was added — "never judged" is
  deliberately distinct from "judged and passed" — and the error-code count pin
  moved 21 → 22 as a deliberate act, not a widening.
- `rust/goldens/` is **frozen**. The TypeScript was the oracle exactly once;
  `capture-ts-oracle.mjs` no longer runs at `HEAD` and says so at the top.
  Re-running it means checking out `7c8e18f`. `PROVENANCE.md` records it.
- `spec/evidence/runs/SUITE-001/5eb727e31986.json` still names
  `tests/org-no-subprocess::…`. Left alone deliberately: it is a recorded
  evidence run — a historical artefact of what a past run measured — not a live
  claim.
- `tests/global-setup.ts` now **throws** if `quoin-core` is unreachable, rather
  than letting a suite skip its way to green.
- `make rust-e2e` now runs five files.
- **`make test` is blocked repo-wide by #430** (missing `quire-rs` sibling
  checkout; it fails before vitest starts). Vitest was run directly:
  **1,114 passed, 9 skipped**. The remaining failures are pre-existing and
  caused by uncloned sibling repositories — three in `bench-tier1.test.ts`
  (`corpus/bounds.py`) and two in `semantic-module-template.test.ts`
  (`spec-objects-business`). Two further files time out only under full-suite
  parallel load and pass in isolation.
- This branch was **rebased onto #448** (`8bab74a`), which landed mid-flight.
  Eleven files conflicted; the merged `quoin-schemas` renderer keeps #448's
  strict `oneOf`-of-string-constants rule (FND-007) and this branch's tagged-
  union rendering, and `render_literal` was dropped in favour of the stricter
  `string_literal`. `src/core/types.ts` and both schema digests were
  regenerated with `make types`, never hand-merged.

## Review cycle: `/rust-review` and `/gap-analysis`, both scoped to this diff

Run as skills over `gh pr diff 450 --name-only` only, not the repo.
`/rust-review` raised six findings; **all six are fixed in this PR** (commit 3),
with no gate weakened — no relaxed lint, no new blanket `allow`, no loosened
pin, no lowered threshold.

| # | Sev | Finding | Disposition |
| - | --- | ------- | ----------- |
| 1 | High | A `quoin.yaml` over 1 MiB or a `.git/config` over 4 MiB was read whole by `src/core/org.ts` and sent; the boundary refuses an over-limit field (`CORE_REFUSED`, exit 2) and `call()` throws, so `quoin write` died where the retained rule degrades (FR-027-AC-5). | Fixed. `readIfPresent` stats before reading and treats an over-limit document as an **absent layer**, the same answer `ConfigService::get` and `org_from_git_config` already give, at the same two numbers. The boundary refusal stays, as defence against a caller that is not this one. Two cases in `tests/core-org.test.ts`, one over the ceiling and one under it. |
| 2 | Med | `render_definition`'s tagged-union guard is `members.iter().all(..)`, vacuously true on an empty slice, so `oneOf: []` rendered `export type X = ;` — invalid TypeScript, from a renderer whose contract is to refuse rather than degrade. | Fixed. `!members.is_empty() &&` added, plus a test at the `$defs` entry point: the existing `an_empty_union_is_refused` only exercises `render_type`, a different arm. |
| 3 | Med | `CoreErrorCode::ProtocolSkew` reused for "a config layer did not contribute its content", which is not a protocol-version disagreement; `src/core/reference.ts:250` emits the same code for the real thing. | Fixed. **`CoreErrorCode::Degraded` → `CORE_DEGRADED` → `Outcome::Partial`** added. No count pin blocked it; the one structural pin, `protocol_skew_is_the_one_code_that_still_carries_a_payload`, was **renamed and extended deliberately** to `the_payload_carrying_codes_are_exactly_the_two_partial_ones` — a third entry there is still a design change. |
| 4 | Med | Three silent-clean paths in `main.rs`, each contradicting that file's own "unjudged is not the same as clean": an unreadable registry re-validated a population of **zero** and reported the home clean; an unreadable module was `continue`d past; and an unreadable registry in `ContractGate::others` turned the duplicate-package and import-resolution rules off silently. | Fixed — all three refuse now (`SemanticContractUnavailable`, `semantic/population-unreadable`). `inspect` was **split** into `refuse`, `as_gate_diagnostics` and `materialise` rather than allowed past `clippy::too_many_lines`. |
| 5 | Low | `FakeHost::tampered` was never assigned and its doc said otherwise, so no unit test asserted `ensure_defaults` re-validates what is installed — deleting that call left every unit test green. | Fixed. Two tests: the refusal, and the re-validation happening at all when nothing is tampered. |
| 6 | Low | `Resolved::degraded` was computed in `service::resolve_documents`, discarded, then re-derived as `!issues.is_empty()`. A schema refusing with an empty issue list would report a substituted default as a clean read. | Fixed. `resolve_org_from_documents` returns an `OrgDocumentReport { degraded, issues }` carrying the flag the machinery set; the payload's `degraded` and the exit status both read it. |

One finding was recorded as "structurally unfixable" — `ContractGate` living in
`main.rs`, untested, because `tc_library_containment.rs` would fail a
`quoin-core/src/host.rs` on the filesystem capability the gate genuinely needs.
**That claim was wrong and is withdrawn.** The independent review pointed out
that `tc_library_containment` audits one crate, and `quoin-modules` — which
already owns `SemanticGate`, `ModuleName` and `ModuleRegistry` — is not audited
by it. The gate moved there. See the independent-review table below.

`/gap-analysis` answered its key question — *did any criterion carried by a
deleted TypeScript test fail to land?* — **clean**. All 37 acceptance criteria
across FR-018 (5), FR-019 (4), FR-023 (5), FR-024 (3), FR-025 (11) and FR-027
(9) appear on a real `/// Trace:` line in a tagged test; NFR-003 declares none.
The unbacked matrix rows and status lies `quire coverage` reports for those FRs
are the repo's pre-existing `TC-0NN` condition — `git grep` for those ids on
`origin/main` returns nothing, so they were never backed there either, and this
PR did not introduce them. The skill's `reviews/YY-MM-DD-<slug>.md` SpecReview
artifact was deliberately **not** minted: the mechanical result is recorded
here instead.

## Public API changes (`src/index.ts`) — the `!` in the title, enumerated

The independent review found these undisclosed. Every change to the package's
public surface, and nothing else changed:

**Removed**

- `parseSourceArg(arg: string): Source` — gone entirely. The source-argument
  grammar now lives in `quoin_modules::Source::parse` behind the boundary, and
  no operation returns a `Source` to TypeScript, so there is nothing for the
  export to return.
- `UNRESOLVED_ORG_MESSAGE: string` — replaced by `unresolvedOrgMessage(): string`.
  The sentence is now the boundary's (`config.unresolved_org_message`), so it
  cannot be a module-load-time constant.

**Renamed** (same behaviour, `plugin` -> `module`, matching the rest of the CLI)

- `installPlugin` -> `installModule`
- `listPlugins` -> `listModules`
- `removePlugin` -> `removeModule`
- the re-exported `InstalledPlugin` type -> `InstalledModule`, and its
  declaration moved from `@agent-ix/ts-plugin-kit` to the generated
  `src/core/types.ts`

**Signatures changed**

- `defaultModulesManifest()`: `MarketplaceManifest` -> `string`. It returns the
  committed YAML **text**. Parsing it here only to re-serialise it for the
  request would put the manifest schema in two places.
- `ensureDefaultModules(home = ixHome(), manifest: MarketplaceManifest = ...): void`
  -> `ensureDefaultModules(home?: string, manifest: string = ..., mode: Mode = "lazy"): EnsureDefaultsPayload`.
  Three changes: `manifest` is now YAML text; `home` defaults to `undefined`
  rather than `ixHome()`, because "the home the host resolves" is now a fact
  the far side answers rather than one this side computes; and the function
  returns the reconcile payload instead of `void`.
- `resolveOrg(repoRoot, options)` — `options` is now the named `OrgOptions`
  interface (same fields), and the **environment it sends is narrowed** to the
  `QUOIN_ENV_BINDINGS` names (review finding 1). Behaviour for callers is
  unchanged: `QUOIN_ORG` was always the only variable read.
- `ResolvedOrg` — was a hand-written `{ org?: string; source: OrgSource }`. It
  is now `Omit<ResolveOrgPayload, "source"> & { source: OrgSource }`, derived
  from the generated type (FR-097). Consequences for callers: `degraded` is
  always present, and `org` may be `null` rather than absent.

**Added**

- `unresolvedOrgMessage()`, `installModule`, `listModules`, `removeModule`,
  `type InstalledModule`.
- From `src/core/org.ts`, the three ceilings this side sends under, exported so
  a test can pin them against the boundary's own: `MAX_ENV_VALUE_BYTES`,
  `MAX_CONFIG_LAYER_BYTES`, `MAX_GIT_CONFIG_BYTES`.

`OrgSource`, `resolveOrg`, `originOrg` and every other export keep their names.
The `!` in the title is **correct**: `parseSourceArg` and
`UNRESOLVED_ORG_MESSAGE` are removed and three functions are renamed, so a
consumer importing them fails to compile.

## Second review cycle: the independent review of this PR

https://github.com/agent-ix/quoin/pull/450#issuecomment-5651210081

Seven findings, each confirmed by the reviewer breaking something and watching
it fail, plus two "plausible" items. **All nine are fixed** (commit 4), with no
gate weakened — no relaxed lint, no new blanket `allow`, no loosened pin, no
lowered threshold, no CI trigger change. Every new test was run against the
unfixed code first; what each said is in the last column.

| # | Sev | Finding | Disposition | Probe against the unfixed code |
| - | --- | ------- | ----------- | ------------------------------ |
| 1 | High | `resolveOrg` sent the **whole** `process.env`. The boundary refuses any value over `MAX_SCALAR_BYTES` (4 KiB) with `CORE_REFUSED`/exit 2, so one unrelated oversized variable killed the command — a regression against the retained TypeScript, which only ever read `QUOIN_ORG`. | Fixed by narrowing, not by raising a ceiling. `declaredEnv` sends only the names in `QUOIN_ENV_BINDINGS`, and drops an over-ceiling value rather than dying on it. Three new cases in `tests/core-org.test.ts`. | With `BIGVAR` at 5,000 bytes exported, `tests/core-org.test.ts`: **9 failed \| 2 passed**. The two new cases against the unfixed `declaredEnv`: **2 failed**, both `Error: quoin-core config.resolve_org exited 2`. |
| 2 | High | `ContractGate::others` silently dropped any registered module whose `manifest.yaml` was not a file, so the duplicate-package and import-resolution rules ran over a shrunken — sometimes empty — population and agreed with themselves. | Fixed in the code, not only in a test: `others` now **refuses**, naming the module and both ways to fix it. A deliberate divergence from the retained TypeScript (which filtered with `existsSync`), written down in the code: unjudged is not the same as clean. `validate_installed`'s `continue` past an absent manifest answers a different question — "this module declares no contract of its own" — and is kept, with the distinction stated. | The two new unit tests: **2 failed \| 8 passed** — `assertion failed: verdict.has_errors()` and `an unjudgeable module is not clean`. The reviewer's exact `.filter(\|_\| false)` probe: **4 failed \| 6 passed**, additionally `a_package_already_claimed_by_an_installed_module_is_refused` ("the duplicate package must be refused") and `an_unreadable_sibling_refuses`. |
| 3 | Med/High | The "structurally unfixable" claim was rejected: `tc_library_containment.rs` audits **one** crate, and `quoin-config`, `quoin-modules` and `quoin-semantic` are unaudited by it. | Claim withdrawn. `ContractGate` moved to `quoin-modules::contract_gate` — the crate that already owns `SemanticGate`, `ModuleName` and `ModuleRegistry` — with ten unit tests against a real temporary home built from the committed fixture and the real `src/semantic` contract tree. Finding 2's fix landed as two of those tests. `main.rs` supplies three paths and nothing else: **486 -> 219 lines**. `quoin-semantic` moved from `quoin-core`'s dependencies to `quoin-modules`'. `tc_library_containment.rs` is still untouched and still passes. | n/a (a structural finding). |
| 4 | Med | TS/Rust ceiling drift was unguarded: narrowing both TS constants to 100000 left 13/13 passing. | Pinned **behaviourally**, with no new op and no source scraping: the test sends `limit + 1` bytes for each of `user_config`, `project_config`, `git_config` and `env_value`, and asserts the boundary's own `CORE_REFUSED` diagnostic reports that same `limit_bytes`. Narrowing the TS number means the boundary accepts; widening it means a different `limit_bytes`. Either way the test fails. | With both constants at 100000: **1 failed \| 14 passed** — `AssertionError: user_config: expected 1 to be 2`. The boundary accepted the over-TS-limit payload, so no refusal existed to read. |
| 5 | Low/Med | Undisclosed public-API changes in `src/index.ts`. | Every one enumerated in the section above, and the `!` in the title confirmed correct. | n/a (a documentation finding). |
| 6 | Low | `render.rs`'s interface arm was changed `\|\|` -> `&&` with nothing disclosing it and no test pinning it in either direction. | Disclosed here and in commit 4's message, and pinned by `the_interface_arm_needs_the_type_and_the_properties`: all three cases — interface when `type: object` **and** `properties`, refusal on `properties` alone, `Record<string, string>` and no `interface` on `type: object` alone. | Reverted to `\|\|`, the test fails on the `type: object`-only case: ``called `Result::unwrap()` on an `Err` value: RenderRefusal { pointer: "#/$defs/X", reason: "an object definition must declare `properties`" }``. |
| 7 | Low | #446 scope item 3 asked for **both** upstream error enums to be guarded. `ModulesErrorCode` got a mapping and a count pin; `ConfigErrorCode` (16 variants, `#[non_exhaustive]`) got neither. | `rust/crates/quoin-core/src/ops/config/taxonomy.rs` adds the deliberate three-group mapping (`BadRequest` / `Refused` / `Io`) and pins `ConfigErrorCode::all().len() == 16` with every code listed by name — a **count**, not a loop over `all()`, which re-runs the same match and agrees with itself (the quoin#443 failure mode). | Adding a 17th variant to `quoin-config`: `the_error_mapping_covers_every_config_code` **FAILED** — "quoin-config gained or lost an error code; map it in core_code deliberately, left: 17, right: 16". |

Both "plausible" items were judged **real** and fixed:

- `ContractGate::inspect` returned CLEAN on an unreadable subject manifest
  while `validate_installed` refused on the identical error. It refuses now
  (`semantic/module-unreadable`). The new unit test failed against the unfixed
  code with "an unjudgeable module is not clean".
- `CoreErrorCode::all()` was hand-maintained with no count pin.
  `the_catalogue_is_every_variant_of_the_enum` pins the eight, by name.

**One new deliberate divergence from the retained TypeScript**, stated rather
than smuggled: a module listed in the registry whose `manifest.yaml` is missing
now makes `ContractGate::inspect` refuse, where `installedSemanticModules` in
the deleted `src/plugins.ts` filtered it out with `existsSync`. Checked against
this machine's real `~/.ix/filament/registry.json`: all ten registered modules
have their manifests, so no working home is affected. A home that would be
affected is a home where the duplicate-package rule was silently off.

## Gates

Re-run at commit 4, on `origin/main` = `8bab74a`, cargo invoked from `rust/`
with an explicit `--target-dir`, toolchain 1.98.1:

- `make rust-gate` — **green** (fmt, clippy `-D warnings`, cargo deny, tests,
  e2e caller, differential harness; 123 cases, 0 differences)
- `make rust-e2e` — **green** (5 files, 31 tests)
- `make lint` — **green** (tsc, 25 src modules with no cyclic component,
  eslint, prettier)
- vitest run directly (`make test` is blocked repo-wide by #430) —
  **1,121 passed, 10 skipped, 4 failed**, every failure pre-existing and caused
  by uncloned sibling repositories: three in `bench-tier1.test.ts`
  (`corpus/bounds.py` absent) and one timeout in `core-exec-e2e.test.ts` that
  passes in isolation under `make rust-e2e`.

No CI workflow was changed; the test workflows remain `workflow_dispatch:` only.

Closes #446.


kreneskyp:
## Independent review — changes only (diff vs `origin/main`)

Read-only worktree, gates re-run from scratch. Where a test claims to catch a regression I **broke the thing and watched it fail, then restored**; everything below marked CONFIRMED was reproduced that way. `git status` clean at the end.

Gates, re-run independently: `make rust-gate` green (fmt, clippy `-D warnings`, cargo deny, tests, e2e caller, differential harness), `make rust-e2e` green (5 files / 27 tests), `make lint` green (tsc, import cycles, eslint, prettier).

---

### CONFIRMED

**1 — High — `src/core/org.ts:112` (with `rust/crates/quoin-core/src/ops/config.rs:195-198`). The whole environment crosses the boundary unbounded; review Finding 1 was fixed for three of the four inputs.**

`resolveOrg` sends `env: stringValuedEnv(options.env ?? process.env)` — every variable in the caller's environment. The boundary refuses any single value over `MAX_SCALAR_BYTES` (4 KiB) with `CORE_REFUSED`, exit 2, and `call()` throws at `src/core/org.ts:169`. Finding 1 taught `readIfPresent` to treat an over-limit *document* as an absent layer; the env map — the least controlled of the four inputs, and the only one the user does not author — kept the fatal shape.

Scenario: `export BIGVAR=$(head -c 5000 /dev/zero | tr '\0' x)` then any `quoin write`. Direct probe returned `CORE_REFUSED ... env_name: BIGVAR, limit_bytes: 4096, observed_bytes: 5000`, exit 2. End-to-end with that variable exported, `tests/core-org.test.ts` goes **9 failed | 2 passed (11)**. Real environments carry 4 KiB+ variables routinely (`LS_COLORS`, `BASH_FUNC_*`, a CI token bundle, a `KUBECONFIG` blob). FR-027-AC-5 says a malformed config never stops an author; an oversized unrelated env var now does.

Suggested shape: drop or truncate an over-limit variable caller-side, the same answer `readIfPresent` gives a document. The boundary refusal stays.

**2 — High — `rust/crates/quoin-core/src/main.rs:269-275`. The semantic gate's population can go empty with every gate green. Finding 4's fix covers *unreadable*, not *empty*.**

`ContractGate::others` builds the population it judges against, then drops any registered module whose `manifest.yaml` is not a file (line 274) — silently. With an empty population the duplicate-package rule and the import-resolution rule are vacuous: `all`/`any` over nothing.

Probe: `.filter(|_| false)` added to that chain. `cargo test --workspace` — all ok. `make rust-e2e` — 5 files, 27 tests, passed. Nothing in the suite notices the gate has been turned off. Restored.

Scenario without tampering: module `alpha` is in `registry.json` but its `manifest.yaml` was renamed or lost. Install `beta` declaring the same `package:` as `alpha` — the duplicate-package diagnostic never fires, because `alpha` was dropped from the population before the rule ran. That is the same "unjudged is not clean" class Finding 4 named, and the same failure mode the ticket calls out ("a check passing over an empty population... this program has shipped that failure four times"). The `.map_err` two lines below refuses loudly for an *unreadable* sibling; line 274 stays silent for an *absent* one. Two answers to the same question in one function.

Root cause is structural: this code lives in the one file the containment audit exempts, and `main.rs` carries **zero** `#[cfg(test)]`. There is no unit test of the gate's own rules — `tc_446_040/041/042` cover clean-home, tampered-module and absent-contract, not the rules themselves. A population floor (`assert!(!roots.is_empty())` where a population is expected, or a test with two installed modules asserting the duplicate fires) would have caught the probe.

**3 — Medium/High — I reject the "structurally unfixable" claim for moving `ContractGate` out of `main.rs`; `rust/crates/quoin-core/src/main.rs` (486 lines) is doing three jobs.**

`tc_library_containment.rs` walks **one crate**: `quoin-core/src`, minus `main.rs`. `quoin-config`, `quoin-modules` and `quoin-semantic` are not audited and name `Filesystem` freely today. `quoin-modules` owns the `SemanticGate` trait; `quoin-semantic` owns every primitive `ContractGate` calls (`SemanticValidators::load`, `read_module_semantic`, `duplicate_package_diagnostic`, `resolve_imports`, `derive_package_manifest`, `write_package_manifest`, `registry_pin`); neither depends on the other, so no cycle blocks `ContractGate` living in a domain crate — where it is unit-testable with `tempfile`, which is what finding 2 is missing. The argument in the PR body rules out `quoin-core/src/host.rs`. It does not rule out "the library".

As structure, that is the same finding: `main.rs` now holds the I/O shell (~90 lines), `HostModules` capability wiring (~120) and `ContractGate` semantic *policy* (~230). The rust-style doc's line is "everything decidable lives in the library so it is unit-testable without spawning a process"; the gate's rules are decidable, and they are the part with no unit tests. The doc comment moving from "four things" to "five things" is the honest record of a containment boundary being spent, not a design. Per the owner's hygiene rule this is a **design** finding (one file, two jobs), not a line count.

By contrast `ops/config.rs` (521 lines, 282 production), `ops/modules/mod.rs` (286) and `ops/modules/tests.rs` (331) are each one concern and read fine — no finding.

**4 — Medium — `src/core/org.ts:24-25`. The two ceilings now live in three places and nothing fails when they diverge.**

`MAX_CONFIG_LAYER_BYTES` and `MAX_GIT_CONFIG_BYTES` exist in `quoin-config`, in `ops/config.rs`, and now again in TypeScript. `the_bounds_are_the_crates_own_ceilings` pins Rust to Rust; nothing pins the TypeScript copies to either.

Probe: narrowed the TS `MAX_CONFIG_LAYER_BYTES` to 100000 and `MAX_GIT_CONFIG_BYTES` to 100000. **13/13 tests in `tests/core-org.test.ts` still pass.** Restored. So the caller can silently start discarding valid 200 KiB config layers — the exact degradation Finding 1's fix was for — with the suite green. The reverse drift (Rust ceiling raised, TS not) is equally unguarded, and `MAX_GIT_CONFIG_BYTES` has no TypeScript-side ceiling test at all. One fact, one place: have the boundary report its ceilings, or assert the TS constants against the schema/Rust values.

**5 — Low/Medium — `src/index.ts`. Undisclosed public-API removals and signature changes.**

`parseSourceArg` removed; `UNRESOLVED_ORG_MESSAGE` const replaced by `unresolvedOrgMessage()`; `defaultModulesManifest()` return type `MarketplaceManifest` → `string`; `ensureDefaultModules` return `void` → `EnsureDefaultsPayload` and its `home` default `ixHome()` → `undefined`. The title carries `!`, but neither "User-visible output" nor "Scope decisions stated explicitly" enumerates any of them, and the body's four listed differences are all CLI-level. A downstream importer reads those sections and finds nothing.

**6 — Low — `rust/crates/quoin-schemas/src/render.rs:86-88`. An undisclosed semantic tightening with no test in either direction.**

The interface-dispatch condition changed from `type == "object" || contains_key("properties")` to `&&`. It is not mentioned in the PR body or in Finding 2's row (which is about `!members.is_empty()`, a different line). I reverted `&&` to `||`: **all 39 `quoin-schemas` tests still pass**, including `tc_1612_a_fresh_render_of_the_rust_types_matches_the_committed_digest`. Restored. So nothing pins which `$defs` shape takes which arm. I judged the change itself **safe** — a `$defs` entry with `properties` and no `"type": "object"` now falls to `render_type`, which refuses by name rather than degrading, and `"type": "object"` without `properties` now renders `Record<string, T>` instead of an empty interface, which is better. It just arrives unannounced and unguarded.

**7 — Low — #446 scope item 3 is half delivered.** "Both upstream enums must be guarded by a pinned `ALL.len()` count." `ModulesErrorCode` has it (`ops/modules/taxonomy.rs:121,156`). `ConfigErrorCode` (16 variants, `#[non_exhaustive]`) has **no mapping and no pin** anywhere in `quoin-core` — the only config op wired takes documents by content and never produces one. Defensible today; it means the guard the ticket asked for is absent when Stage 9 wires `config.set`/`config.get`.

---

### PLAUSIBLE (not demonstrated reachable)

- **`main.rs` `ContractGate::inspect`** returns `SemanticVerdict::default()` (CLEAN) via `let Ok(result) = quoin_semantic::read_module_semantic(...) else` when a manifest is unreadable/unparsable/not a mapping, while `HostModules::validate_installed` maps the identical error to `ModulesError::SemanticContractUnavailable` and refuses. Two answers to one condition in one file. I traced `install.rs:158` → the gate runs after `materialize`; on the ad-hoc path `declared_name` is `None` so the manifest was already parsed, and on the reconcile path `ensure_defaults` later runs `validate_installed`, which does catch it. Effectively closed today, fragile by construction.
- **`CoreErrorCode::all()`** is hand-maintained with no count pin and no compiler exhaustiveness check. Pre-existing shape, but this diff adds a variant to it.

---

### Probed and found clean (so the empty half of this list is explicit)

- **No gate weakened.** `tc_library_containment.rs` byte-identical. No `clippy::too_many_lines` allow — `inspect` was genuinely split into `refuse`/`as_gate_diagnostics`/`materialise`. No CI trigger change (no `push:`/`pull_request:` added to a test workflow). The one new `#[allow]` in `main.rs` is single-lint, single-site, with a `reason =`.
- **The payload-carrying pin was strengthened, not loosened.** `the_payload_carrying_codes_are_exactly_the_two_partial_ones` asserts an exact two-element vec; the old pin asserted one code. No case the old pin covered was dropped.
- **`src/core/org.ts` document selection is genuinely anti-narrowing.** Broke it: 5 tests fail, including the population floor `expect(reached).toBe(3)`. Three documents each as the only org source in its own real tree, plus a worktree/`commondir` case. Not a self-written fixture.
- **`src/core/modules.ts` key-order restoration is guarded.** Reordered the nested `semantic: ["package","semanticCore","exports"]` list: `tests/core-modules.test.ts` fails. It pins against the real `registry.json` written by a real install, with `toHaveLength(1)` as a floor. It carries no content pre-filter, so #448's `--json` key-order class does not apply to it.
- **Deleted property tests are genuinely restated.** `tc_446_source_arg_properties.rs` and `tc_446_org_url_properties.rs` replace fast-check generation with exhaustive loops carrying explicit floors (`26 * 26 * 5`, `sections.len() == 64`) — arguably stronger than what they replaced. Read each deleted file at `origin/main` and matched criterion by criterion.
- **All 37 ACs across FR-018/019/023/024/025/027 have a real `Trace:` line** and the spot-checked carriers measure their criterion rather than restating it. FR-025-AC-7's amendment is honest and has two real carriers.
- **Every new `MAX_*` bites before the allocation.** All `check_bound` calls in `ops/config.rs` and the ceilings in `ops/modules/wire.rs` precede any work, and `MAX_REQUEST_BYTES` (64 MiB) exceeds every field ceiling, so the field bounds are reachable rather than shadowed. Boundary cases tested at limit and limit+1.
- No `todo!`/`unimplemented!`/`dbg!`/stub file introduced anywhere in the diff.

---

Blocking, in my view: **1** and **2**. **3** is the structural reason **2** had nowhere to be tested.

kreneskyp:
All seven findings and both "plausible" items are fixed in commit `60a3198` on this branch. No follow-up PR; nothing split out. No gate weakened — no relaxed lint, no new blanket `allow`, no loosened pin, no lowered threshold, no CI trigger change. Every new test was run against the unfixed code first, and each of the three probes you named was re-applied.

**1 (High) — the whole environment crossed the boundary.** Narrowed, not raised. `declaredEnv` now sends only the names in `QUOIN_ENV_BINDINGS` (today: `QUOIN_ORG`), and drops a value over `MAX_ENV_VALUE_BYTES` rather than dying on it — the answer the retained TypeScript gave, which only ever read `QUOIN_ORG`.

Your probe re-applied: with `BIGVAR` at 5,000 bytes exported, `tests/core-org.test.ts` was **9 failed | 2 passed**. The two new cases against the unfixed `declaredEnv` were **2 failed | 13 passed**, both `Error: quoin-core config.resolve_org exited 2`. The direct boundary probe named it exactly: `{"code":"CORE_REFUSED","context":{"env_name":"BIGVAR","field":"env_value","limit_bytes":"4096","observed_bytes":"5000","op":"config.resolve_org"}}`, exit 2. Fifteen pass now.

**2 (High) — the silent drop over an empty population.** Fixed in the code, not only in a test. `others` now refuses when a registered module has no `manifest.yaml`, naming the module and both ways to fix it, so the population is the registry's or there is no verdict. This is a deliberate divergence from the retained TypeScript — `installedSemanticModules` filtered with `existsSync` — and it is written down in the code and in the PR body. `validate_installed`'s `continue` past an absent manifest answers a different question ("this module declares no contract of its own") and is kept, with the distinction stated in a comment.

Against the unfixed gate the two new unit tests failed with `assertion failed: verdict.has_errors()` and `an unjudgeable module is not clean` (**2 failed | 8 passed**). Your exact `.filter(|_| false)` probe took it to **4 failed | 6 passed**, additionally `a_package_already_claimed_by_an_installed_module_is_refused` ("the duplicate package must be refused") and `an_unreadable_sibling_refuses`.

**3 (Med/High) — "structurally unfixable" was wrong.** Withdrawn. You are right that `tc_library_containment.rs` audits one crate. `ContractGate` now lives in **`rust/crates/quoin-modules/src/contract_gate.rs`** — the crate that already owns `SemanticGate`, `ModuleName` and `ModuleRegistry`, and which needs only free functions from `quoin-semantic` (the reverse direction would make the policy crate depend on the installer). Ten unit tests there, against a real temporary home built from the committed fixture and the real `src/semantic` contract tree; finding 2's fix landed as two of them. `main.rs` supplies the three paths and nothing else: **486 → 219 lines**. `quoin-semantic` moved out of `quoin-core`'s dependencies into `quoin-modules`'. `tc_library_containment.rs` is untouched and passes.

`ops/config.rs`, `ops/modules/mod.rs` and `ops/modules/tests.rs` left alone as you said.

**4 (Med) — unguarded ceiling drift.** Pinned behaviourally, with no new op, no source scraping and no schema regeneration: the test sends `limit + 1` bytes for `user_config`, `project_config`, `git_config` and `env_value`, and asserts the boundary's **own** `CORE_REFUSED` diagnostic reports that same `limit_bytes`. Narrow the TS number and the boundary accepts (exit 1, no refusal to read); widen it and `limit_bytes` disagrees.

Your probe re-applied — both TS constants at 100000: **1 failed | 14 passed**, `AssertionError: user_config: expected 1 to be 2`.

**5 (Low/Med) — undisclosed API changes.** Enumerated in the PR body under "Public API changes (`src/index.ts`)": `parseSourceArg` removed; `UNRESOLVED_ORG_MESSAGE` → `unresolvedOrgMessage()`; `installPlugin`/`listPlugins`/`removePlugin` → `installModule`/`listModules`/`removeModule`; `InstalledPlugin` → `InstalledModule` with its declaration moving to the generated `src/core/types.ts`; `defaultModulesManifest()` `MarketplaceManifest` → `string`; `ensureDefaultModules` gaining a `mode`, taking YAML text, defaulting `home` to `undefined` and returning `EnsureDefaultsPayload` instead of `void`; `ResolvedOrg` derived from the generated payload (so `degraded` is always present and `org` may be `null`); the three ceiling constants newly exported. The `!` is right: removals plus three renames break a consumer at compile time.

**6 (Low) — the undisclosed `||` → `&&`.** Disclosed in commit 4's message and the PR body, and pinned by `the_interface_arm_needs_the_type_and_the_properties`, which asserts all three cases rather than only the changed one. Reverted to `||` it fails on the `type: object`-only case: ``called `Result::unwrap()` on an `Err` value: RenderRefusal { pointer: "#/$defs/X", reason: "an object definition must declare `properties`" }``.

**7 (Low) — #446 scope item 3 half delivered.** `rust/crates/quoin-core/src/ops/config/taxonomy.rs` adds the deliberate three-group mapping (`BadRequest` / `Refused` / `Io`, with a `_` arm that exits Internal and *says* `mapping: unrecognised`) and pins `ConfigErrorCode::all().len() == 16` with all sixteen listed by name. A count, not a loop over `all()`. `ops/config.rs` became `ops/config/mod.rs` to hold the sibling; its content is unchanged apart from the `mod` line.

Probe: adding a 17th variant upstream, `the_error_mapping_covers_every_config_code` **FAILED** — "quoin-config gained or lost an error code; map it in core_code deliberately, left: 17, right: 16".

**Both plausible items were real.**

- `ContractGate::inspect` returning CLEAN on an unreadable manifest while `validate_installed` refused on the identical error: it refuses now (`semantic/module-unreadable`). The new test failed against the unfixed code with "an unjudgeable module is not clean".
- `CoreErrorCode::all()` hand-maintained with no count pin: `the_catalogue_is_every_variant_of_the_enum` pins the eight, by name.

**Gates**, re-run at commit 4 on `origin/main` = `8bab74a` (already the base; nothing to rebase onto after the fetch), cargo from `rust/` with an explicit `--target-dir`, toolchain 1.98.1:

- `make rust-gate` — green (123 differential cases, 0 differences)
- `make rust-e2e` — green (5 files, 31 tests)
- `make lint` — green
- vitest directly — 1,121 passed, 10 skipped, 4 failed, all four pre-existing (three `bench-tier1` on the absent `corpus/bounds.py`, one `core-exec-e2e` timeout that passes in isolation)

Commit 1 re-verified reversible to a tree identical to `8bab74a` (`git diff --cached --stat 8bab74a` empty).

Not merged — yours to land.

Issue #457, opened 2026-09-13
Stage 4 cutover: wire quoin-change-assurance through quoin-core dispatch and delete src/change-assurance/


kreneskyp:
Parent: #373. Successor to #455. Traces FR-101, NFR-025.

Wire `quoin-change-assurance` into the `quoin-core` binary and delete the
1,782 lines of TypeScript it replaces. **A port is not a cutover** — #455
delivers a crate; this ticket makes it the code that runs.

## Blocked on decisions, now made

- **#394** — durability: two named primitives, parent-directory fsync stays.
- **#406** — `__proto__` is refused on both sides, with a stable error code.
- **#408** — the 33 non-canonical store files are frozen; the replay gate reads
  them as a recorded allowance, not as passes.

## Scope

1. Register `change_assurance.*` as command-shaped operations in
   `rust/crates/quoin-core/src/dispatch.rs`. The unit of IPC is
   `quoin-core <domain>.<op>`, never a bare function.
2. Every op gets a `pub const MAX_*_BYTES` applied **before any host is
   consulted**, and `tc_412_the_transport_ceiling_stays_above_every_domain_bound`
   gains its entries.
3. stdout = payload, stderr = canonical JSON diagnostics. No diagnostic on
   stdout, ever. Full exit taxonomy `Ok 0 / Partial 1 / Refused 2 / Invalid 3 /
   Internal 4`, with the crate's error codes mapped onto it deliberately and the
   mapping pinned by a count.
4. Rewire the 9 files under `src/commands/change-assurance/` through
   `src/core/exec.ts` — `intake.ts`, `receipt.ts`, `recover.ts`, `schema.ts`,
   `seal-record.ts`, `seal-attestation.ts`, `verify-receipt.ts`, `common.ts`,
   `index.ts`. **This is its own commit, demonstrated revertible.**
5. **Then** delete `src/change-assurance/*.ts` and its tests in the same commit.
   `src/store/schemas/*.json` is the vendored contract and does **not** go.

## Route coverage, not handler coverage

The trap #447 found: a unit test calling the handler function cannot see a
swapped match arm — the drift guard, the unit tests and difftest all stay green
while `quoin change-assurance <cmd>` breaks for a user.

**Every `OPERATIONS` entry must be handed to the real binary by its wire
spelling**, by an integration test asserting a payload only that handler writes.

## Tests that must be split, not deleted

`tests/change-assurance-schema-assets.test.ts` (67) asserts the schema assets
[...]

PR #459, opened 2026-09-13
feat(rust): port change-assurance to a quoin-change-assurance crate (#455)

### LINKED ISSUE #455: [Stage 4] quoin-change-assurance: records, attestations, receipts and verification
Port `src/change-assurance/` — **1,782 lines, 6 files** — to a Rust crate
`rust/crates/quoin-change-assurance/`.

Parent: #373. Traces FR-100, NFR-025. **No TypeScript is deleted in this ticket**
— cutover and deletion are a separate ticket, per FR-101.

## Why this one first

It is the only genuinely leaf domain left in the program. #425 established that
the ordering rule is **type closure**, not the import graph, so this ticket
states its closure up front:

| Needs | Owner | Status |
|---|---|---|
| canonical JSON, JCS, strict parse | `quoin-store` | **merged** |
| blake3 record digest, sha256 file digest, digest domains | `quoin-store` | **merged** |
| durable + content-addressed atomic write, staging recovery | `quoin-store` | **merged** |

Nothing else. After #376 broke the cycle, `src/change-assurance/` imports **zero**
evidence symbols — its shared surface moved down into `src/store/`, which
`quoin-store` already covers in full. No YAML, no subprocess, no network, no
`process.env`.

## Scope

| TS file | Lines | Rust module | Note |
|---|---:|---|---|
| `records.ts` | 593 | `records` | seal/verify, lineage, ix-flow chain hashing |
| `verify.ts` | 553 | `verify` | `verifyChangeAssurance`, `verifyReceipt` |
| `types.ts` | 226 | `model` | record / attestation / receipt / decision |
| `store.ts` | 215 | — | **mostly already in `quoin-store`**; only `intakeAttestation`'s directory staging and `recoverChangeAssuranceStaging` are new here |
| `attestations.ts` | 145 | `attestations` | seal/verify/bytes |
| `index.ts` | 50 | `lib.rs` | |

## Constraints

**The library half names no host capability.** Storage goes through
`quoin-store`. The staging/intake flow is the one filesystem seam and sits
behind a trait with a disk and an in-memory implementation and **one** analysis
over both — the shape `quoin-validators` established with
`RepoSource`/`DiskRepo`/`MemoryRepo`.

**One canonical form in the program, not two that agree by coincidence.**
`hashIxFlowEvent` (`records.ts:466-480`) carries a *third* JSON serializer of
its own (`records.ts:500-510`), separate from both JCS and the pretty writer.
It must call `quoin-store`'s function. If a byte comparison proves the two
differ, that difference is a **finding to report**, not a port detail to
reproduce.

**Durability semantics are decided, not rediscovered.** Follow #394: mutable
named records (`records/<digest>.json`) use `write_atomic`; the parent-directory
fsync stays; directory-granular attestation publish stays a composition here,
not `quoin-store` surface.

**Digest surface.** blake3 over JCS bytes with the top-level `digest` member
removed, 64 lowercase hex unprefixed (`src/store/integrity.ts:194-197`).
sha256 for the ix-flow chain with genesis `"0".repeat(64)` (`records.ts:464`).
Use `quoin-store`'s `CanonicalDigest` / `RawFileSha256Digest` newtypes; do not
carry bare `String` digests.

## House rules

`#![forbid(unsafe_code)]`; one `thiserror` error enum at the crate boundary with
stable codes; newtypes over bare `String` for every id; clippy clean at
`-D warnings`; `/rust-review` is the style guide and quoin's
`.claude/skills/rust-style/SKILL.md` outranks it. Avoid giant files — one module
per concern, `mod.rs` re-exporting.

Invoke cargo from `rust/` with an explicit `--target-dir`; toolchain 1.98.1.

## Acceptance

- Seal, verify, lineage, receipt and ix-flow chain results **byte-identical** to
  TypeScript on a golden corpus captured once from TS. No live Node oracle
  (FR-101 AC-5).
- Every criterion carried by `tests/change-assurance.test.ts` (973 lines) is
  restated on a Rust test with a **criterion-level** `Trace:` tag
  (`FR-100-AC-n`), never a bare `FR-100`.
- No self-fixture tautology: a criterion is not carried by a test asserting
  against a fixture the test itself wrote, nor by a check over an empty
  population.
- `make rust-gate` green.

kreneskyp:
Closes #455.

Ports `src/change-assurance/` (1,782 lines, 6 files) to a new crate,
`rust/crates/quoin-change-assurance`. **No TypeScript is deleted here** —
cutover and deletion are #457.

## Lines of Rust added

| | files | lines |
|---|---|---|
| `src/` | 20 | 6,156 |
| `tests/` | 7 | 2,101 |
| **Rust total** | **27** | **8,257** |
| `oracle/capture-change-assurance-oracle.mjs` (one-time capture, not in the test path) | 1 | 720 |
| `tests/fixtures/oracle.json` (the committed capture) | 1 | — |
| `DIVERGENCE.md` | 1 | 76 |

## Module layout

```
src/lib.rs                 crate docs: the integrity boundary, and where the rules live
src/error.rs               one thiserror enum, stable QCA-* codes, structural refusal kinds
src/ids.rs                 a newtype per identity and per digest domain
src/model/{json,outcome,record,attestation,receipt,decision}.rs
                           the four sealed shapes, the closed 35-reason vocabulary,
                           UTF-16 ordering, and the retained ix-flow event
src/records.rs             sealChangeRecord / verifyChangeRecord / verifyLineage / validateDecision
src/attestations.rs        sealAttestation / verifyAttestation
src/verify/{input,audit,lineage,proofs,mod}.rs
                           verifyChangeAssurance and the receipt it always produces
src/intake/{mod,memory,disk}.rs
                           trait EvidenceStore + MemoryEvidenceStore + DiskEvidenceStore,
                           one analysis over both
```

`intake/disk.rs` is the only module that names a host capability
(`std::fs`, the process id, the clock — the last two only inside a staging
directory's name). Paths, atomic and content-addressed writes, canonical
bytes, strict parse, blake3 and sha256-of-a-file all come from `quoin-store`;
nothing is reimplemented. The genuinely new storage work is
`intake_attestation`'s directory staging (a pair is published whole by one
rename) and `recover_staging`, which `quoin-store`'s file-granular recovery
cannot do.

## The `hashIxFlowEvent` serializer question — measured, and they agree

`hashIxFlowEvent` (`records.ts:467`) carries a **third** JSON serializer,
`canonicalIxFlowJson`. This crate does not port it: it hashes `quoin-store`'s
JCS bytes.

The capture records, for each of 6 events, the sha256 `hashIxFlowEvent`
produced **and** the bytes `canonicalizeJcs` produced. The events were chosen
to separate the two if anything can: an astral character, every escape JSON
defines, `-0`, `1e21`, `1e-7`, `9007199254740991`, an empty object and array,
and keys crossing the surrogate boundary where UTF-16 and scalar order
disagree.

**Result: byte-identical on all six, and the hashes match.**
`canonicalIxFlowJson` cannot be distinguished from JCS on anything
`parseStrictJson` admits, so there is no third serializer in the port and no
divergence to report. Pinned by
`tc_455_the_ix_flow_serializer_is_jcs_byte_for_byte` and
`tc_455_every_captured_event_hashes_to_what_the_oracle_hashed`.

## Criteria restated, and where

Every test carries a criterion-level `/// Trace:` tag. 49 tests: 9 unit,
39 integration, 1 doctest.

| Criteria | Where |
|---|---|
| FR-063-AC-1..8 | `tests/tc_455_records.rs` |
| FR-063-AC-9 | `tests/tc_455_intake.rs` |
| FR-063-AC-10 | `tests/tc_455_ix_flow.rs` |
| FR-063-AC-11 | `tests/tc_455_boundary.rs` |
| FR-064-AC-1, 2, 4, 5, 7 | `tests/tc_455_attestations.rs` |
| FR-064-AC-3, 6, 8 | `tests/tc_455_intake.rs` |
| FR-064-AC-9 | `tests/tc_455_boundary.rs` |
| FR-065-AC-1..12 | `tests/tc_455_verify.rs` |
| FR-100-AC-3 (refusal classification parity) | `tests/tc_455_verify.rs` |
| FR-100-AC-8 (no host capability in the library half) | `tests/tc_455_boundary.rs` |
| FR-100-AC-9 (interrupted publish, orphan recovery) | `tests/tc_455_intake.rs` |
| FR-100-CON-4 (one canonicalizer, one digest) | `tests/tc_455_boundary.rs`, `tests/tc_455_ix_flow.rs` |

The oracle capture is committed and **no test runs Node** (FR-101 AC-5). Every
census-style test carries an anti-vacuity floor, and every expectation compared
against was produced by the TypeScript, never by this crate.

Reason coverage: **34 of the 35 reasons** in the closed vocabulary are reached
by receipts this crate produced, asserted as an exact figure by
`tc_455_thirty_four_of_the_thirty_five_reasons_are_reached`.

## Two faithfulness bugs the oracle caught, fixed here

Both were found by the parity tests, not by reading:

1. `schemaInvalidReceipt` wraps a non-object record **twice** before digesting
   it (`{retained_record: {retained_record: raw}}`). The port wrapped once and
   produced a different `record_digest`. Reproduced, with the reason in a
   comment: the digest is the only name an unreadable record has.
2. `verifyLineage`'s first statement re-verifies the subject record, so a
   record whose digest does not verify fails the **lineage** check as well,
   classified `parent_mismatch`. The port skipped it and reported lineage
   valid.

## Findings against the retained TypeScript (not "fixed" here)

`DIVERGENCE.md` records four. The one worth reading:

**`parent_missing` is unreachable as a receipt reason.** The lineage check is
built with `check(reasons, "invalid")` while `parent_missing` sits in the
*incomplete* precedence set, so `verifyChangeAssurance` throws
`verification receipt outcome disagrees with reason precedence` while
validating the receipt it just built. Scenario `genesis-with-a-parent` in the
capture is that throw. The port reproduces the refusal exactly, and
`tc_455_refusals_happen_where_the_oracle_refuses` asserts refusal parity on
every scenario.

The others: `exact()` names undeclared members in map order rather than source
order; `Revision`/`size_bytes` are `u64`, so an integral double beyond 2^63 is
refused where the oracle accepts it; and a retained attestation with a
malformed `retained_output.digest` makes the oracle throw while building the
receipt, while this crate emits `null` and still returns one — the argument for
each is in `DIVERGENCE.md`.

## Gate output

```
$ make rust-gate
rust-gate: fmt, clippy -D warnings, cargo deny, tests, the end-to-end caller and the differential harness passed

$ make lint
check-import-cycles: 24 src modules, no cyclic component
$ prettier --check .
All matched files use Prettier code style!

$ cargo test -p quoin-change-assurance
9 + 5 + 4 + 5 + 3 + 8 + 14 passed; 0 failed   (+ 1 doctest)

$ QUOIN_CORE=… pnpm exec vitest run
Test Files  1 failed | 93 passed | 1 skipped (95)
     Tests  3 failed | 1059 passed | 10 skipped (1072)
```

The 3 vitest failures are `tests/bench-tier1.test.ts` reaching for `corpus/`
and `python3`, which a worktree does not carry — the same environmental class
as the known baseline failures on main, and no TypeScript changed in this PR.

Nothing was weakened to make anything pass: no pin loosened, no lint relaxed,
no blanket `allow`, no test deleted, no gate lowered.


kreneskyp:
## Review by measurement — gates re-run independently, all green

Re-ran every gate in a clean worktree at `8b9d6a7` with an explicit `--target-dir`:

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — 0 warnings
- `cargo test -p quoin-change-assurance` — 48 tests + 1 doctest, all passing
- `cargo deny check` — advisories, bans, licenses, sources ok

Measured against the source rather than the summary:

- **Host capability** — `std::fs` and `std::process` appear in `intake/disk.rs` and nowhere else in `src/`. `tc_455_only_the_disk_store_names_a_host_capability` enforces it with a ≥14-file floor.
- **Trace tags** — 39 tests, 39 criterion-level tags, every one `-AC-n` or `-CON-n`. Zero bare `FR-100`. No `status_lies` shape.
- **Anti-vacuity** — all four boundary censuses carry a floor. The one in `tc_455_no_second_canonicalization_or_digest_implementation_exists_here` asserts the dependency set exactly, which is the right way to make a second implementation loud rather than quiet.
- **One canonicalizer** — confirmed: no `blake3::`, no `serde_json` in this crate; the single `Sha256` is in `model/decision.rs` for the ix-flow event hash. The JCS byte-parity measurement over 6 adversarial payloads is the right answer to the third-serializer question.

## One divergence the owner should see — recorded, not blocking

`DIVERGENCE.md` §4 cases 2 and 3: when a retained attestation carries a malformed `retained_output.digest` or `attestation_digest`, the TypeScript **throws while building its own receipt**; this crate emits `null` for the unparseable digest and still returns a receipt.

That is an observable change at cutover — `Internal` (exit 4) becomes `Ok`/`Partial` — not merely a different message. The reasoning given is sound and I agree with it: one malformed member inside one retained attestation should not destroy the verification of every other proof in the record, and the receipt is the only artifact that would have recorded why. The receipt still cites the attestation by its selection digest, so the trail is intact. No such record exists in any retained store.

It is a deliberate loosening of a refusal, which is the class this program normally treats as a finding, so it is called out here rather than left only in the file. It goes on #457's cutover checklist as a difftest case.

## Findings against the retained TypeScript — correctly recorded rather than fixed

`parent_missing` is unreachable as a receipt reason: the lineage check is built `invalid` while `parent_missing` sits in the incomplete precedence set, so `verifyChangeAssurance` throws validating the receipt it just built. Reproduced exactly, and the reason census asserts 34 of 35 as an exact figure rather than a floor. That is the right handling — the TypeScript is deleted at cutover, so no retrofit.

## Minor, non-blocking

- `src/model/record.rs` is 1,078 lines, the largest file in the crate. Under the "avoid giant files" rule it is the one worth splitting next time it is touched; not worth churning this PR for.
- `oracle/capture-change-assurance-oracle.mjs` is marked binary by git, so it is undiffable in review. It is deleted by #457 under FR-101 AC-5, so it does not need fixing.

No gate weakened, no lint relaxed, no test deleted, no pin touched. Merging.

Issue #470, opened 2026-09-13
[Stage 6] quoin-jsonschema: one ajv-parity validator for quoin-semantic and quoin-measurement


kreneskyp:
Wave 3 of the Stage 6 measurement port. Ports `intervention-schema.ts` (195) and `operational-schema.ts` (4), and extracts `SchemaValidator` / `SchemaError` / `SchemaKeyword` from `quoin-semantic/src/schema.rs:180-282` into a new `quoin-jsonschema` crate shared by `quoin-semantic` and `quoin-measurement`. Vendors `intervention-experiment-v1.schema.json` and `operational-evidence-v1.schema.json`.

Gate: `quoin-semantic`'s existing tests pass unchanged through the re-export; the vendored intervention schema is byte-equal to `canonicalJson(interventionExperimentSchema)` from a one-time `.mjs` capture; the vendored operational schema is byte-equal to `src/measurement/schemas/`; ajv-to-`jsonschema` verdict parity on a named corpus; `should_validate_formats` plus the two `date-time` formats wired.

This wave also closes #440: capture the TypeScript schema, apply exactly two recorded deltas (RFC-3339 case widening, `blake3:` removal per #409), and assert those are the only differences.

Trace: FR-100-AC-4, FR-098. Related: #392, #403, #434.

Parent: #373

kreneskyp:
Landed in PR #485, merged `7598293`. `rust/crates/quoin-jsonschema` exists on `main` with its `src`, `tests`, vendored `schemas`, one-time `oracle` capture and `DIVERGENCE.md`.

`quoin-semantic/src/schema.rs` became a re-export shell over it, so there is exactly one `SchemaValidator` type in the workspace — `quoin-semantic`'s existing `tc_378_03x` tests pass unchanged through the re-export, which was this ticket's gate. `tc_470_parity.rs`'s divergence register is now the pattern every later parity wave copies.

Left open by oversight when the PR merged; the work is done.

PR #483, opened 2026-09-13
[Stage 6 W2] quoin-measurement record types and the two pure record renderers (#469)

kreneskyp:
Closes #469. Wave 2 of Stage 6 (#373).

**Rebased onto `main` @ 055f5b3 (#468 / PR #482 merged).** What the rebase
changed is at the bottom, under "Rebase onto #468".

## What is here

525 TypeScript lines ported into the new `quoin-measurement` crate:
`intervention-types.ts` (122), `operational-types.ts` (136),
`intervention-report.ts` (132), `operational-report.ts` (135).

**No TypeScript is deleted.** All four source files are untouched.

These four sources import only each other, so this half of the crate needs
`serde`, `serde_json` and `thiserror` and no workspace crate. (`quoin-store`
and `quoin-yaml` are in the manifest for #468's half.)

| module | lines |
|---|---:|
| `src/common/{identity,literal_bool,mod,producer,recorded_evidence,scalar,wire_enum}.rs` | 137, 52, 18, 45, 37, 238, 113 |
| `src/intervention/{agent_eval,intake,mod,record,report}.rs` | 72, 98, 13, 255, 281 |
| `src/operational/{github_release,mod,record,report}.rs` | 58, 12, 498, 211 |

Largest module of this wave 498 lines, under both the 500 soft and 700 hard
ceilings that #468's `tc_468_module_sizes` enforces. `src/lib.rs` is shared and
is #468's file with this wave's three modules added.

## Gate results

All cargo invocations from `rust/` with an explicit
`--target-dir .worktrees/469-measurement-records/rust/target`.

| gate | result |
|---|---|
| rendered text byte-identical to `renderInterventionReport` / `renderOperationalReport` on committed fixtures | **green** — 15 cases |
| `InterventionIntakeError` code set is a `Copy` enum with `as_str`/`from_code` + round-trip tests | **green** |
| `cargo clippy --workspace --all-targets -- -D warnings` | **clean** |
| `cargo test --workspace` | **green** — 907 passed, 0 failed, 2 ignored |
| `make rust-gate` | **green** — fmt, clippy, `cargo deny` (advisories/bans/licenses/sources ok), tests, `rust-e2e`, `rust-difftest` 12 cases / 0 differences |

`make rust-gate` needed `make install` in the worktree for its `rust-e2e` and
`rust-difftest` legs; that installed `node_modules` only and left
`pnpm-lock.yaml` unchanged.

## The oracle is a file, not a runtime

[...]

Issue #468, opened 2026-09-13
[Stage 6] quoin-measurement base: types, plans, profiles, validation, store seam and raw-evidence accounting


kreneskyp:
Wave 1 of the Stage 6 measurement port. Ports `types.ts` (99), `compare.ts` (113), `validate.ts` (217), `date-time.ts` (50), `plans.ts` (103), `profiles.ts` (77), `store.ts` (117), `atomic-file.ts` (50) — 826 lines — into the new `quoin-measurement` crate.

Owns `MeasurementPlan`, `MeasurementCollection`, `MeasurementObservation`, `MeasurementComparison`, `AssuranceProfileSummary`, `MeasurementSource`.

Needs `quoin_store::{canonical_json, store::store_root, store::write_content_addressed, digest_file_sha256, RawFileSha256Digest}` and `quoin_yaml::from_str`.

Includes the `raw_evidence` module (`rawEvidenceFor`, `verifyRawEvidenceReferences`, `assertGoverningDefinition` from `intervention.ts:115-179`), which is what cuts the `operational → intervention` edge.

Gate: the 48 retained collections under `spec/evidence/measurements/` read then re-serialize byte-identical through `quoin_store::canonical_json_bytes`; `MeasurementSource`/`DiskMeasurement`/`MemoryMeasurement` with one analysis over both; the two `date-time` grammars disagree on `...t...z` by assertion; the census names 48 and fails at 0.

Trace: FR-100-AC-2, FR-100-AC-4, NFR-025-AC-3.

Parent: #373

kreneskyp:
Two fidelity defects in this wave's type model were found by #473 (wave 6) and fixed in PR #490, not absorbed. Noting them here because this is the ticket that shipped them.

1. **Unmodelled `population` members were dropped on re-serialisation.** `DIVERGENCE.md` §3.3 as landed said the typed parse drops them and that *'nothing in the corpus carries one'*. The second half was false: **25 of the 14,644** observations under `spec/evidence/measurements/` carry `exclusions` and `namedMisses`. `MeasurementPopulation` now keeps every unmodelled member verbatim and flattens it back out. §3.3 is amended rather than left standing next to the correction.

2. **An absent `dimensions` was flattened to `{}`.** **305 of 14,644** observations state none, and absent vs `{}` are different bytes in the report JSON. A `Dimensions` newtype now models the distinction.

[...]

Issue #464, opened 2026-09-13
[Quoin#464] Workspace module-size census: 24 source files over 500 lines, 7 over 700


kreneskyp:
Measured on `origin/main`: 179 source files across 17 crates. 24 exceed 500 lines, 7 exceed 700.

| lines | file |
|---:|---|
| 1,443 | `quoin-schemas/src/render.rs` |
| 1,078 | `quoin-change-assurance/src/model/record.rs` |
| 875 | `quoin-store/src/replay.rs` |
| 782 | `quoin-semantic/src/manifest.rs` |
| 751 | `quoin-config/src/service.rs` |
| 732 | `quoin-store/src/digest.rs` |
| 732 | `quoin-core/src/dispatch.rs` |

`digest.rs` and `dispatch.rs` earn it (domain documentation and self-census machinery respectively). `render.rs` and `record.rs` are the two that most need splitting.

They cleared the bar because nothing checks. Add a `tc_NNN_module_sizes` census — `include_str!` over the crate's own sources, 700-line ceiling, anti-vacuity floor on module count, allow-list entries each carrying a one-line reason.

Not a port blocker. New Stage 6 crates carry the test from the start.

kreneskyp:
Partial: `quoin-change-assurance/src/model/record.rs` (1,078) is being split as a third commit on the #457 cutover PR, and both #457 and #458 are adding the `tc_NNN_module_sizes` census to their crates. `quoin-schemas/src/render.rs` (1,443) and the rest remain open here.

PR #485, opened 2026-09-13
feat(rust): extract quoin-jsonschema and vendor the measurement schemas (#470)

### LINKED ISSUE #409: [Quoin#409] intervention-schema admits a blake3: digest where its sibling digest field is sha256:-only
Found while landing `quoin-store` (#380, PR #405), which makes the three digest domains non-substitutable **types**: `RawBytesDigest` (blake3, bare hex), `CanonicalDigest` (blake3 over JCS bytes, bare hex), `RawFileSha256Digest` (sha256, `sha256:<hex>`). No `From`, no `Into`, no shared trait — because nothing in a hex string tells you which question it answers.

`src/measurement/intervention-schema.ts` contradicts that in one file:

```ts
const digest = { type: "string", pattern: "^sha256:[a-f0-9]{64}$" };
const immutableVersion = {
  type: "string",
  pattern:
    "^(v?[0-9]+[.][0-9]+[.][0-9]+([-+][0-9A-Za-z.-]+)?|[a-f0-9]{40}|(sha256|blake3):[a-f0-9]{64})$",
};
```

`digest` accepts `sha256:` only. `immutableVersion`, two declarations later, accepts a semver, a 40-hex git sha, **or** `sha256:`/`blake3:`. So the same document admits a `blake3:`-prefixed value in one slot and refuses it in the other, with no statement anywhere of what a `blake3:` immutable version would denote or who writes one.

## What this ticket wants

Answer, from the producers rather than from the schema:

1. **Has any `blake3:` value ever been written to an `immutableVersion`?** The #380 replay counted 42,320 `sha256:` references across 59 stores and no `blake3:` prefixed reference in a stored record. If the answer is none, the alternative is unexercised and the schema is describing a capability nobody has.
2. **If none: drop the `blake3:` alternative.** A schema alternative with no producer is not flexibility, it is an unowned branch that the Rust port must either implement or refuse — and either choice is then a behaviour change nobody asked for.
3. **If some exist: say what the digest is over.** `blake3` is the algorithm for both `RawBytesDigest` and `CanonicalDigest`, which digest different bytes. A bare `blake3:<hex>` in a version slot does not distinguish them, so it cannot be verified — which is the exact failure mode the type split in `quoin-store` exists to prevent.

Do this before the measurement stage (Stage 6) ports `intervention`. After cutover the Rust behaviour is the behaviour, and a refusal added then looks like a port defect rather than the deliberate narrowing it would be.

Related: #406, #407, #408 (the other #380 findings). Digest-domain background: `rust/quoin-store/COMPATIBILITY.md`.

### LINKED ISSUE #440: isRfc3339DateTime duplicates the shared parseRfc3339DateTime, and the two already disagree on case
`src/measurement/intervention.ts:347` declares a private `isRfc3339DateTime`:

```ts
function isRfc3339DateTime(value: string): boolean {
  const match = /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.\d+)?(?:Z|([+-])(\d{2}):(\d{2}))$/.exec(value);
  …
  return month >= 1 && month <= 12 && day >= 1 && day <= daysInMonth(year, month) &&
         hour <= 23 && minute <= 59 && second <= 59 && offsetHour <= 23 && offsetMinute <= 59;
}
```

with its own `daysInMonth` and its own leap-year rule at line 376. `src/measurement/date-time.ts` already exports `parseRfc3339DateTime`, which performs the same checks with its own copy of the same month table and the same leap-year rule, and which `operational.ts` and `github-release-operational.ts` both consume.

Two implementations of one predicate, in the same module directory, one of them exported for exactly this purpose.

### Why this is worth a ticket rather than a shrug

They agree **today**. That is the whole problem: nothing makes them agree tomorrow, and nothing would report it if they stopped.

#436 was the third copy of this predicate, and it had already drifted — it left every range check to `Date.parse`, so it accepted `2026-02-30` and rolled it to March 2, and the rolled number then decided a reported status. The drift was invisible for as long as nobody compared the three side by side. These two are the remaining pair, and `CLAUDE.md` states the rule: *"two readers over the same files drift, and the second gets written by whoever needs a field the first does not expose."*

### Differences that exist right now

`isRfc3339DateTime` returns a boolean; `parseRfc3339DateTime` returns the parsed number or `null`. The intervention copy therefore does not carry the final `Date.parse` sanity check the shared reader ends with, and its regex requires an uppercase `T` and `Z` where the shared one accepts `[Tt]` and `[Zz]`. So they are not merely redundant — they already disagree on case, and a value accepted by one surface is refused by the other.

Which spelling is correct is a real question and should be answered once, in one place, rather than settled differently by each caller that happens to need it.

### Fix

Consume `parseRfc3339DateTime` from `intervention.ts` and delete the private copy and its `daysInMonth`, resolving the case question explicitly rather than by whichever implementation survives. #436 is the shape to follow: keep a caller's own stricter surface where it is deliberate, delegate only the ranges, and run the test against the unfixed code first so the change is shown to do something.

### Found by

Sizing `parseAssuranceArgument` for the Rust port (#384). Enumerating every RFC-3339 implementation in the tree turned up three; #436 fixed the loose one, and this is the duplicate left behind. Deliberately not folded into #436 — that ticket is about a validator that accepted impossible dates, and this is about a predicate declared twice.

kreneskyp:
Wave 3 of Stage 6 (epic #373). Extracts a new crate `quoin-jsonschema` and vendors the two measurement schemas into it. **No TypeScript was deleted** — this is a port wave, not a cutover.

Closes #440. Closes #409.

## What landed

1. **New crate `rust/crates/quoin-jsonschema`.** `SchemaValidator`, `SchemaError`, `SchemaErrorParams`, `SchemaKeyword` and `identity_counts` move here verbatim from `quoin-semantic/src/schema.rs`. `quoin-semantic` re-exports them, so its callers and its ten `tc_378_03x` unit tests are unchanged (byte-identical `mod tests`).
   - `SchemaValidator::compile` still resolves in `quoin-semantic` and still returns `SemanticError`, via a `SchemaValidatorExt` extension trait — so there is exactly one `SchemaValidator` type in the workspace, not two.
2. **`src/measurement/intervention-schema.ts` (195 lines) and `operational-schema.ts` (4 lines) ported**, with both JSON schemas vendored into `crates/quoin-jsonschema/schemas/`.
3. `VendoredSchema` / `VendoredValidator` / `ValidDocument` — parse, don't validate: a validated document returns a newtype, modelled on `RawFileSha256Digest::parse_stored`.

## The two owner-ruled deltas

The vendored intervention schema differs from the TypeScript by **exactly two** edits, both recorded in `crates/quoin-jsonschema/DIVERGENCE.md` and asserted as an enumerated set:

| # | Pointer | TypeScript | Vendored | Ruling |
|---|---------|-----------|----------|--------|
| 1 | `/$defs/rfc3339DateTime/pattern` | strict `T` / `Z` | permissive `[Tt]` / `[Zz]`, fractional seconds | RFC 3339 §5.6; the permissive grammar in `src/measurement/date-time.ts:2` wins. Widening cannot break a retained record. |
| 2 | `/$defs/immutableVersion/pattern` | `(sha256\|blake3):[a-f0-9]{64}` | `sha256:[a-f0-9]{64}` | #409 — no producer, no instance; all 8 repo hits are doc comments, guards and rejection fixtures. |

Delta 1 is wired to **the** RFC-3339 grammar landed by #468 (`quoin_measurement::date_time::Rfc3339DateTime`) — no second grammar was written.

**The operational schema keeps its `blake3:` alternative** — it is vendored byte-for-byte and the ruling was scoped to the intervention schema.

## Gate

| # | Criterion | Result |
|---|-----------|--------|
| 1 | `quoin-semantic` tests pass unchanged through the re-export | **Pass, with one disclosed deviation** — see below |
| 2 | Vendored operational schema byte-equal to the TS source | Pass — `cmp` clean; asserted by `tc_470_the_operational_schema_is_vendored_byte_for_byte` |
| 3 | Vendored intervention schema differs from the one-time capture by exactly the two deltas | Pass — `tc_470_the_intervention_schema_carries_exactly_the_two_recorded_deltas` enumerates every JSON-pointer difference and asserts the **set** equals the two deltas with exact before/after text |
| 4 | ajv-to-`jsonschema` verdict parity on a named corpus | Pass — 125 documents, see below |
| 5 | `should_validate_formats` + both `date-time` formats wired | Pass — `tc_470_formats.rs` asserts from both sides |
| 6 | `cargo clippy --workspace --all-targets -- -D warnings` | Pass, clean |
| 7 | `cargo test --workspace` | Pass |
| 8 | `make rust-gate` | Pass (`EXIT=0`) |

### Gate 1 — the one deviation, stated plainly

`rust/crates/quoin-semantic/tests/golden_parity.rs` has **one line changed**: the `use` line gained the identifier `SchemaValidatorExt` (1 insertion, 1 deletion). `SchemaValidator::compile` needs that trait in scope. No assertion, expectation, fixture, or behaviour changed, and no test was deleted, split or skipped. Every other `quoin-semantic` test file, including the ten in-module `tc_378_03x` tests, is byte-identical.

### Gate 4 — the parity corpus

**125 documents**, chosen as *3 retained base records × a fixed, named mutation ladder*, not sampled or generated:

- Bases: `spec/evidence/interventions/quoin-270-cli-eval-sentinel-contract.json`, and records 0 and 1 of `spec/evidence/operational/pairs/c1b30a18….json`.
- Ladder: each required key removed; an unknown key added; both `const`s broken; the identity pattern broken; 8 `observed_at` forms; 7 `producer.tool_version` forms; `environment: {}`; a bare-hex digest; `raw_evidence: []`; an empty owner.
- Verdicts: 30 accepted / 95 refused overall — intervention 8/37, operational 22/58. Both schemas carry both verdicts, asserted with a floor of 100.

The oracle is ajv 8.20.0 run **once** through `oracle/capture-ajv-verdicts.mjs`, reproducing both retained ajv configs verbatim (`intervention.ts:22-27`, `operational.ts:30-35`). Verdicts are contractual; error text, order and count are not.

**Four declared divergences** (ajv verdict ≠ Rust verdict), each listed in `DECLARED_DIVERGENCES` with its ticket, and each asserted to have actually been exercised: three lowercase-`t`/`z` `observed_at` forms (#440, delta 1) and one `blake3:`-prefixed `tool_version` (#409, delta 2). Every other one of the 121 documents agrees exactly.

## Concurrency with PR #483

**`quoin-measurement` was not touched at all** — no `Cargo.toml` line, no `src`, no tests, no module-tree change. `quoin-jsonschema` sits *below* `quoin-measurement`, so it cannot build-depend on it; the `date-time` check is injected as `fn(&str) -> bool` into `VendoredSchema::compile`, and the **tests** take a dev-dependency on `quoin-measurement` (cargo permits a dev-dependency cycle) so the real grammar is provably what gets wired. `tc_470_boundary.rs` asserts the manifest keeps that shape.

## Census tests (in the new crate only)

- `tc_470_module_sizes` — 700 hard / 500 soft, **empty** allow-list, anti-vacuity source floor. Carries no `Trace:` tag by design: it is a #464 engineering criterion with no owning requirement.
- `tc_470_boundary` — no second RFC-3339 grammar, hasher or canonicalizer (#463/#464); no module reads a schema off the filesystem; the manifest holds no grammar dep and no build cycle.

## Unverified / not claimed

- `node scripts/check-trace-tags.mjs` reports 3 unresolved criterion references, all in `tests/core-bundle-snapshot.test.ts`. I checked these are **identical on unmodified `main`** — pre-existing, not introduced here.


kreneskyp:
## Review — measured, not read

Approved. Merging.

**Gates, re-run independently at `9e72090` in my own worktree**
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test --workspace` — green.
- TypeScript touched: **zero files**. Correct for W3; this wave extracts and vendors, it does not cut over.
- Largest module `src/validator.rs` at **374 lines**; every other module under 230. Comfortable margin, unlike W2's 498.

**The three disclosures, checked rather than taken**

1. **The `golden_parity.rs` deviation is exactly what the PR says it is.** `git diff main...9e72090` on that file is one `use` line: `SchemaValidatorExt` added so the trait resolves. No assertion, fixture, or expectation moved, nothing deleted or split. The ten `tc_378_03x` tests are byte-identical in the re-exported `schema.rs`. This is a move, and it reads as one.

2. **Both rulings were measured against the actual tree, and I re-ran the measurement.** `grep -rl "blake3:" spec/evidence tests/fixtures corpus` → nothing. A scan for a lowercase `t` date separator under `spec/evidence/` → nothing. So #409's narrowing refuses a value that has never been written, and #440's widening accepts a form nothing currently stores. **No retained verdict in this repository moves in either direction.** That is what makes the narrowing in §2 acceptable: a narrowing is normally the dangerous direction, and it is only safe here because the population is empty and the type system already refuses to mint the value.

3. **The operational schema keeping its `blake3:` alternative is the right call.** It is vendored byte-for-byte and no ruling was taken about it. Closing that gap belongs to the operational wave with its own measurement, not to this one by side effect. Scoping a ruling to what you measured is correct.

**What makes the parity claim real rather than asserted**
- `CORPUS_FLOOR = 100` — an anti-vacuity floor on a 125-document corpus.
- `tc_470_the_corpus_carries_both_verdicts_for_both_schemas` — a corpus that refused everything would agree perfectly and prove nothing. This forbids that.
- Every `DECLARED_DIVERGENCES` entry is asserted to **have actually fired**. A declared divergence that stops diverging fails the test, and an undeclared one fails it too. That is the shape that keeps a divergence register honest.
- The capture records its ajv version, both configs, and the quoin revision it ran at.

**The one-grammar claim is enforced, not documented.** `quoin-measurement` is a **dev**-dependency, and `tc_470_the_manifest_holds_no_grammar_and_no_cycle` reads the manifest and fails if it ever becomes a build dependency, or if `chrono`/`time`/`sha2`/`blake3`/`serde_yaml` appear. Passing the `date-time` check in as `fn(&str) -> bool` is what lets this crate sit below `quoin-measurement` without either duplicating the grammar or forming a cycle. That is the right structure, and it is the reason §1's ruling can't quietly re-fork.

**Oracle (FR-101 AC-5):** frozen. No `Command::new`, node, pnpm or tsx anywhere in `src/` or in any `.rs` test — the `"node": "24.15.0"` hits are provenance fields inside the captured goldens. The two `.mjs` capture scripts under `oracle/` ran once and are inert. **They are deleted by the measurement cutover's commit 2 (#479), with the TypeScript** — recorded here so it rides with the deletion rather than becoming its own ticket.

**Not claimed as passed, correctly:** `check-trace-tags.mjs`'s 3 unresolved references in `tests/core-bundle-snapshot.test.ts`, verified identical on unmodified `main`. Pre-existing, not this PR's.

Closes #470, #440, #409.

Issue #471, opened 2026-09-13
[Stage 6] quoin-measurement interventions: intake, validation and the agent-eval producer


kreneskyp:
Wave 4 of the Stage 6 measurement port, concurrent with W5. Ports `intervention.ts` (423) and `agent-eval-intervention.ts` (235) — 658 lines.

Needs `quoin_evidence::adapters::agent_eval::parse_agent_eval`, `quoin_store::digest_file_sha256`, `quoin_jsonschema::SchemaValidator`.

Gate: records under `spec/evidence/interventions/` round-trip byte-identical; record-id path encoding (`intervention.ts:33-45`, the `p-` / `b-` disjoint namespaces) tested at the 128-byte cap; `resolveRawPath`'s escape refusals (`intervention.ts:176-200`) tested; the #407 symlink divergence declared (`quoin_store::digest_file_sha256` refuses what `readFileSync` digests — `digest.rs:325-332`).

Resolves #409 and declares #407.

Trace: FR-100-AC-4, FR-100-AC-9.

Parent: #373

kreneskyp:
**Required deliverable added at PR #483 review (2026-09-13): collapse the three raw-evidence representations to two.**

`quoin-measurement` now carries one concept at three trust levels, each honestly documented but not yet reconciled:

| type | where | trust level |
|---|---|---|
| `raw_evidence::RawEvidenceReference` | #468 | validated newtypes (`RawEvidencePath`, `NonEmptyText`, `RawFileSha256Digest`), no serde |
| `raw_evidence::RawEvidenceClaim` | #468 | unchecked, bare `String`s |
| `common::recorded_evidence::RecordedEvidenceReference` | #469 | serde-read, holder newtypes (`EvidencePath`, `MediaType`, `Digest`) that hold without validating |

Two is the right number — as-recorded and verified — and `RawEvidenceClaim` vs `RawEvidenceReference` is already that split. The third exists because #469 needed serde and was scoped out of validation, which was the correct call for that wave; `common/identity.rs:113` says so explicitly and names this ticket as the destination.

This wave is where it converges, because this is where records are actually parsed. Collapse to two representations and assert it: a test that fails if a third type describing a retained evidence file appears in the crate. Do not resolve it by leaving the third in place with a comment.

Issue #466, opened 2026-09-13
[Quoin#466] Retained workflow exports may be multi-document YAML streams


kreneskyp:
`quoin_yaml::from_str` refuses a multi-document stream (`rust/crates/quoin-yaml/src/lib.rs:70-75`). `src/measurement/github-release-operational.ts:27` parses whole GitHub Actions workflow files, which may contain `---` separators.

Population is **unmeasured**. Stage 6 W5 greps its corpus for `^---$` before declaring parity; an empty population is a valid result.

Not a port blocker.

kreneskyp:
## Measured during quoin#472 (Stage 6 W5): an empty population

The operational port needed the answer, so the scan is now committed as
`rust/crates/quoin-measurement/tests/tc_472_github_release.rs::tc_472_no_retained_workflow_export_is_a_multi_document_stream`.

**Result:** **0** `^---$` document markers, over **1** retained workflow export
(`spec/evidence/github-actions/quoin-271-release-v0.22.5-workflow.yml`). The
whole `spec/evidence/github-actions/` tree holds 4 files, one of which is YAML.

That is an **empty population, not a clean bill.** With one export in the
corpus the scan cannot tell whether a multi-document stream would be
mishandled — it only establishes that none exists today. So this issue should
stay open on the underlying question:

- `quoin_yaml::from_str` reads the **first** document of a stream and drops the
  rest silently. `github-release-operational.ts` and the Rust port both look up
  `workflow.jobs` and `workflow.on` on that first document.
- A retained export that is a stream would therefore be read partially, with no
  refusal, and the producer's "exactly one configured release job" check would
  measure the wrong document.

The committed scan means the next retained workflow export is **measured**
rather than assumed: a multi-document export added to the corpus fails the test
with a message pointing back here. What #466 still needs is a decision on what
the reader should do about a stream — refuse it, or read all documents — and
that decision is not settled by a corpus of one.

Provenance: quoin#472, PR #487.

Issue #478, opened 2026-09-13
Stage 6 wiring: expose the measurement surface through quoin-core dispatch (no deletion)


kreneskyp:
Wave 11 of the Stage 6 measurement port. Exposes the measurement surface through `quoin-core` dispatch. **No TypeScript deleted in this ticket** — that is the cutover (#479, #480).

Twelve `OPERATIONS` entries, following the `assurance` build/render precedent in `dispatch.rs:22-43`:

    measurement.record
    measurement.produce_agent_eval_intervention
    measurement.produce_github_release_operational
    measurement.build_report    / measurement.render_report
    measurement.build_comparison
    measurement.build_series
    measurement.build_portfolio / measurement.render_portfolio
    measurement.build_graph_portfolio / measurement.render_graph_portfolio

**Route coverage is not handler coverage (#447).** `measurement.record` is one route with three write paths, so it needs three integration tests handing the real binary its wire spelling and asserting a payload only that handler writes — the returned path is `<store>/measurements/<id>.json` vs `<store>/interventions/p-<id>.json` vs `<store>/operational/<id>.json`, which distinguishes all three.

`OPS_SOURCES` in `dispatch.rs:337-341` gains `("measurement", "ops/measurement/mod.rs", include_str!(...))` plus one entry per file under it, or `the_bound_census_can_see_every_ops_module` fails by design.

Twelve `MAX_*_BYTES` consts, every one entered in `tc_412_the_transport_ceiling_stays_above_every_domain_bound` and applied before any host is consulted: `MAX_COLLECTION_BYTES`, `MAX_INTERVENTION_RECORD_BYTES`, `MAX_OPERATIONAL_RECORD_BYTES`, `MAX_PRODUCER_DEFINITION_BYTES`, `MAX_ASSURANCE_DOCUMENT_BYTES`, `MAX_WORKFLOW_YAML_BYTES`, `MAX_RETAINED_EXPORT_BYTES`, `MAX_PORTFOLIO_ROOTS_BYTES`, `MAX_GRAPH_MAPPING_BYTES`, `MAX_METRIC_NAME_BYTES`, `MAX_REVISION_BYTES`, `MAX_RECORD_ID_BYTES` (128 ASCII).

Do **not** add `quoin_store::MAX_DIGESTED_FILE_BYTES` (256 MiB) to that list — it is a file bound in `quoin-store`, not an `ops::*` request bound, and `tc_412` enumerates `ops::*` only.

[...]

Issue #386, opened 2026-09-12
[Quoin#386] [Stage 6] measurement: the record, store and report subsystem


kreneskyp:
Port the connected core of `src/measurement/` — 24 files, 5,734 lines minus the graph crate — to `rust/quoin-measurement`.

## The planned five-way split does not build. Do not use it.

#373 proposes `measurement-core`, `-portfolio`, `-graph`, `-operational`, `-intervention`. Measured against the import graph, that split creates **three value-level Cargo cycles** and omits 2,117 lines entirely.

The three fatal cycles, all value edges in both directions — no amount of moving type declarations dissolves them:

1. **core ↔ intervention** — `report.ts:8` imports `readInterventionRecords` from intervention; `intervention.ts:19` imports `loadMeasurementPlans` from plans and `intervention.ts:15` imports `writeFileAtomicNoReplace` from atomic-file.
2. **core ↔ operational** — `report.ts:14` imports `readOperationalRecords`; `operational.ts:14,15` import `writeFileAtomicNoReplace` and `parseRfc3339DateTime`.
3. **core → operational → intervention → core** — `report.ts:14` → `operational.ts:16` → `intervention.ts:19`.

Two further cycles (core ↔ portfolio, core → graph → portfolio → core) come from putting `index.ts` inside a lower crate; those dissolve if the facade is hoisted.

**Root causes, both of which the directory listing hid:**
- `report.ts` is a *top*-of-stack consumer that pulls both record stores. Putting it in "core" is what makes core depend on its own dependents.
- `operational` and `intervention` are **not peers**. `operational` depends on `intervention` in four places (`operational.ts:16,20`, `github-release-operational.ts:7`, `operational-types.ts:1`). They are one layered subsystem.

The file graph itself is a clean DAG with no cycles. The cycles exist only in the proposed boundaries.

## What to build instead

**One crate.** If maximum granularity is later wanted, the acyclic layering the code supports is `base` (types, validate, plans, store, compare, profiles, date-time, atomic-file — 825) → `records` (intervention + operational, merged — 2,206) → `reporting` (report + portfolio, merged — 755) → facade (index.ts — 136). Those cuts are real but each produces an 800–2,200 line crate with a single consumer: Cargo bookkeeping without isolation. Start as one crate and split only if a second consumer appears.

`index.ts` must be the crate root / facade, never a module inside a lower layer.

## Seams
[...]

PR #487, opened 2026-09-13
[Stage 6 W5] Port operational intake and the GitHub-release producer (#472)

### LINKED ISSUE #472: [Stage 6] quoin-measurement operational evidence: the pair writer, the store-wide lock and the GitHub release producer
Wave 5 of the Stage 6 measurement port, concurrent with W4. Ports `operational.ts` (531) and `github-release-operational.ts` (295) — 826 lines.

Needs `quoin_yaml::from_str`, `quoin_jsonschema`, the crate's `raw_evidence` module and an injected `Clock`.

Gate: records under `spec/evidence/operational/pairs/` round-trip byte-identical; the directory lock's mutual exclusion asserted by two threads, not commented; the 10-second deadline behind the injected clock; the `on:` key survives `quoin_yaml` (already covered by `quoin-yaml/src/lib.rs:170`); a `^---$` scan of retained workflow exports confirms single-document.

Answers #404 for this domain and preserves #440's two grammars.

Trace: FR-100-AC-4, FR-100-AC-9.

Parent: #373

kreneskyp:
Ports `src/measurement/operational.ts` (531 lines) and
`src/measurement/github-release-operational.ts` (295 lines) into the existing
`quoin-measurement` crate under `src/operational/`.

**Not a cutover.** Zero TypeScript deleted, zero `.ts` files touched. The
retained tree stays the oracle until quoin#479.

**3,512 lines of Rust added, 94 removed**, across 11 new modules, 4 new test
files and a fifth digest domain in `quoin-store`.

## What the gate measured

Real numbers, from `rust/` with `--target-dir rust/target-472`:

| gate | result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | clean, zero warnings |
| `cargo test --workspace --locked` | **951 passed, 0 failed** |
| `cargo deny check` | advisories ok, bans ok, licenses ok, sources ok |
| largest new module | `operational/semantics.rs`, **304** lines (soft ceiling 500) |
| `tc_468_module_sizes` over-soft list | still empty |

`make rust-e2e` and `make rust-difftest` were **not** run: both need a pnpm
install and a TypeScript build inside the worktree, and nothing in this wave is
reachable from the `quoin-core` boundary — `quoin-measurement` is not wired
into its dispatch by this PR. Everything else in `make rust-gate` was run and
is in the table above.

### The producer reproduces the retained pair byte for byte

`tc_472_github_release` hands the Rust producer the four committed files under
`spec/evidence/github-actions/` in a scratch repository holding nothing else,
and requires the same content-addressed file name
(`c1b30a18…acf8.json`) and the same bytes as the pair the retained TypeScript
produced. It passes. The oracle is the committed pair, not a fixture the test
wrote.

### The four things the plan asked for

1. **Pairs round-trip byte-identically.** `tc_472_operational_pairs`: every
   file under `spec/evidence/operational/pairs/` through
   `quoin_store::canonical_json_bytes`. Anti-vacuity floor asserted — the
   retained count is 1, and the test fails at 0. A second test validates both
   records in each pair against the vendored `operational-evidence.v1` schema.
2. **The lock's mutual exclusion, by two threads.** `tc_472_operational_lock`:
   two threads race for one store root on a real filesystem, 20 rounds each,
   and the test fails if a second writer is ever inside. quoin#386 asked for
   this. The 10-second deadline is behind an injected `Clock`, so the refusal
   test advances virtual time rather than sleeping; `SystemClock` is what
   production passes and it is `Date.now()` plus a 5 ms wait, unchanged.
3. **`on:` survives `quoin_yaml`.** Asserted against the retained workflow
   export, both directions: the `on` key is present and no `true` key is. Under
   YAML 1.1 — every libyaml binding — the plain scalar `on` resolves to the
   boolean `true` and the producer would never find a trigger again.
4. **The `^---$` scan (quoin#466).** Zero document markers across 1 retained
   workflow export. Reported on #466 as what it is: an **empty population**, not
   a clean bill. The scan is committed, so the next retained export is measured
   rather than assumed.

## The three-to-two collapse, and who did it

`quoin-measurement` carried one concept at three trust levels.
`RawEvidenceClaim` (quoin#468, bare `String`s) existed only because quoin#469's
serde-read `RecordedEvidenceReference` did not exist yet. This wave parses the
records, so the claim is deleted and `verify_raw_evidence_references` takes the
recorded reference directly — one fewer conversion at every intake.

`tc_472_evidence_representations` reads the crate's own `src/` and fails if a
third type declaring all four members of a retained-evidence pointer ever
appears, and a second test fails if the deleted name comes back in code.

**quoin#471 carries the same obligation and had not landed it when this branch
was cut. It was done here.** #471 should drop it from its own scope and keep
both waves' `pub mod` lines when it rebases.

## Divergences

All declared in `rust/crates/quoin-measurement/DIVERGENCE.md` §11. In short:
an unreadable retained record refuses with a code rather than a bare `Error`;
the producer's two failure kinds are two variants rather than a caught
exception; `linked` compares `Subject`/`OperationalScope` structurally rather
than as canonical-JSON strings; findings sort in Unicode scalar order rather
than UTF-16 code-unit order; schema finding *messages* are the validator's own,
not ajv's; `write_content_addressed` links and fsyncs where the retained code
renames; and the lock's deadline is injected.

## Defects found

**quoin#466 measured, not closed.** The multi-document scan found zero markers
over one export, which is an empty population — it cannot settle whether a
multi-document stream would be mishandled. Reported there.

Nothing else was found that this PR absorbed. No pin weakened, no lint relaxed,
no gate lowered; `tc_468_boundary` caught a forbidden token in a doc comment
during this work and the comment was reworded, not the gate.

Closes #472.


kreneskyp:
## Review — measured, not read

Approved. Merging. #488 (W4) rebases onto this.

**Gates, re-run independently at `1f5b8e6` in my own worktree**
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean.
- `cargo test --workspace --locked` — green, **112 test binaries**.
- TypeScript touched: **zero files**. Correct for W5.
- Largest new module 304 lines (`operational/semantics.rs`). The over-soft list stays empty.

**The lock is what #386 asked for, and that is the headline.** Two real threads, 20 rounds each, racing for one lock on a real filesystem, failing if two are ever inside — with the 10-second deadline on an injected `Clock` so the refusal test advances a counter instead of sleeping. The reasoning in the module header is right: a test that slept for the deadline would be deleted by the first person it annoyed. `operational.ts:296-318` described its exclusion in prose and asserted none of it; this asserts it.

**The fifth digest domain is correctly shaped.** `RecordFileNameSha256` answers a genuinely fourth question — not "what bytes are in this file" and not "what value is this record", but "what do I call the file". It is minted only by `digest_record_file_name`/`digest_record_pair_file_name`, stored bare, and deliberately has **no `to_stored`** because a basename carries no algorithm label. The NUL join means `(a, b)` cannot collide with `(a\0b, "")`. This extends the non-substitutable-domain model rather than bending it.

**Two things about the evidence, stated plainly rather than rounded up**

1. **The round-trip corpus is one pair.** `RETAINED_PAIRS = 1` is the repo's actual population, and the floor is honest and documented as append-only. But it means the byte-identity claim does not rest mainly on that walk — it rests on `tc_472_github_release`, which hands the ported producer four committed GitHub exports in a scratch repo and gets the committed pair back **byte for byte, same content-addressed file name**. That is a real oracle: the retained artifact, not a fixture the test wrote. Worth knowing which of the two tests is carrying the weight.

2. **The `^---$` scan found 0 markers over 1 export — an empty population, and you said so.** Reporting that on #466 rather than booking it as a clean result is exactly right. #466 stays open on the real question, which is what the reader should do with a multi-document stream, not whether one happens to exist today.

**The three→two collapse landed here.** `RawEvidenceClaim` deleted, and `tc_472_evidence_representations` is a source census with a floor, a comparison against a *named* set (so finding two different types fails as loudly as finding three), and a separate guard that the deleted name has not returned in code. #488 did the same collapse concurrently — it drops its copy in the rebase, and keeps both waves' `pub mod` lines.

**One finding, recorded rather than fixed, because the obvious fix is worse.** `json_bridge::to_serde` duplicates `quoin_core::ops::change_assurance::support::to_serde` — same route, `canonical_json_bytes` then `from_slice`, differing only in error type. The natural home would be `quoin-store`, which owns both `JsonValue` and the canonicaliser. **It is not, and should not be:** `quoin-store` has no `serde_json` dependency and that absence is deliberate. Adding one so the store can convert to a type it exists to avoid would cost more than six duplicated lines. `from_serde` has no counterpart anywhere and is genuinely new. The module header already names the precedent it follows, which is the honest way to carry this.

`to_iso_string` is likewise right: an inverse of the one grammar, placed in the one calendar module, existing because the retained pair carries `.000Z` and dropping the milliseconds would change the bytes of every record this producer writes. Not cosmetic.

**Also correct:** `tc_468_boundary` caught a forbidden token in a doc comment and the comment was reworded rather than the gate touched. That is the rule working.

**Not run, and declared:** `make rust-e2e` / `make rust-difftest` — nothing here is reachable from the `quoin-core` boundary yet. W11 (#478) is where that becomes measurable.

Closes #472.

Issue #407, opened 2026-09-12
[Quoin#407] rawEvidenceFor digests whatever a symlink points at — and quoin-store will refuse it


kreneskyp:
Found by #380 (PR #405), which hardened the Rust side and therefore created a behaviour difference that has to be resolved deliberately rather than discovered at cutover.

## Today

`rawEvidenceFor` (`src/measurement/intervention.ts:115`) is a bare `readFileSync` with no guards. It will follow a symlink out of the tree and digest whatever it lands on, and it will happily read a device node or a file being written underneath it.

## In `quoin-store`

`digest_file_sha256` refuses:

- symlinks
- non-regular files
- inputs over 256 MiB
- files whose length changed during the read

## Why this needs a decision rather than a merge

**This can make a previously-digesting input refuse.** That is the whole point of filing it. An input that produced a digest yesterday produces a refusal after cutover, and unless someone has decided that is correct, it reads as the port having broken something.

The guards are right. A digest of "whatever the symlink pointed at when we happened to read it" is not an identifier — it is a value that depends on the filesystem at read time, which is exactly the property a retained digest must not have. The length-changed check is the same argument: a digest taken over a file being concurrently written identifies nothing.

But "the guards are right" is not the same as "nothing was relying on the old behaviour". Before Stage 4 cuts over:

1. Decide whether any producer legitimately points `rawEvidenceFor` at a symlink. If one does, that producer is the defect, not the guard.
2. Apply the same guards to the TypeScript so the refusal appears **before** the cutover, not as part of it. A behaviour change bundled into a language change cannot be attributed to either.
3. Record the refusal in the retained record rather than throwing it away — a refusal to digest is evidence about the input, and FR-018's rule is that a non-success state is part of the contract.

Related: #375 owns the `cargo deny` and workspace rulings; this one is producer-side and belongs to whoever owns `src/measurement/`.

Issue #486, opened 2026-09-13
[Quoin#486] Retained intervention record's file name predates the p-/b- namespacing


kreneskyp:
## What

`spec/evidence/interventions/quoin-270-cli-eval-sentinel-contract.json` is not at the path `interventionPath` would choose for it.

`src/measurement/intervention.ts:39-45` names a record's file `p-<record_id>.json` (portable ids) or `b-<base64url>.json` (ids containing `/` or `:`). The retained record has no prefix — it was published before that encoding landed.

## Why it matters

Reads are unaffected: both the retained TypeScript and the quoin#471 Rust port list the directory rather than resolving a name. But **re-publishing that record writes a second file beside it rather than over it**:

```
spec/evidence/interventions/quoin-270-cli-eval-sentinel-contract.json    # the retained one
spec/evidence/interventions/p-quoin-270-cli-eval-sentinel-contract.json  # what a re-run publishes
```

Both would then be read, as two records with the same `record_id`.

## How it was found

`tc_471_the_producer_reproduces_the_retained_record` (quoin#471) runs the ported producer on the retained definition and the two retained agent-eval reports in a copy of the tree. The **bytes** it publishes are byte-identical to the retained record; the **path** is the `p-` one.

## Fix

`git mv` the retained file to `p-quoin-270-cli-eval-sentinel-contract.json`. Its bytes do not change, so no digest over the record's content moves.

The quoin#471 test asserts both halves — the produced path is the `p-` one, and the unprefixed name no longer exists — so the rename fails that test and points at `quoin-measurement/DIVERGENCE.md` §11.8, which is where to drop the expectation.

Found while porting for quoin#471; not absorbed there, because the store is the thing that is wrong.

PR #488, opened 2026-09-13
[Stage 6 W4] quoin-measurement interventions: intake, validation and the agent-eval producer

### LINKED ISSUE #471: [Stage 6] quoin-measurement interventions: intake, validation and the agent-eval producer

kreneskyp:
Ports `src/measurement/intervention.ts` (423 lines) and `src/measurement/agent-eval-intervention.ts` (235 lines) into the existing `quoin-measurement` crate under `src/intervention/`.

**Not a cutover.** Zero TypeScript deleted, zero `.ts` files touched. The TS remains the oracle until quoin#479.

## What the gate measured

Real numbers, from `rust/` with `--target-dir` inside the worktree.

| gate | result |
| --- | --- |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean, exit 0 |
| `cargo test --workspace` | exit 0 — 111 test binaries, 0 failures |
| `make rust-gate` | exit 0 — fmt, clippy, `cargo deny` (advisories/bans/licenses/sources ok), tests, `rust-e2e` (47/47 vitest), `rust-difftest` (12 cases, **0 differences**) |

**2751 lines of Rust added**, 92 deleted (2884 / 92 including `DIVERGENCE.md`). Largest new module: `intervention/agent_eval/produce.rs` at 382 lines — every new module is under the 500-line soft ceiling, and `tc_468_module_sizes` (hard 700, soft 500, empty allow-list) is unchanged and passing. No second census was added.

## The strongest result

`tc_471_the_producer_reproduces_the_retained_record` copies the retained `spec/` tree into a tempdir, deletes the retained intervention record, runs the ported producer on the retained definition and the two retained agent-eval reports, and gets the retained record back **byte for byte**. That is an oracle the port did not write.

## The ticket's gate items

- **Byte-identical round-trip** — `tc_471_corpus.rs` walks `spec/evidence/interventions/`, asserts the count (1), and fails at 0 rather than passing by measuring nothing.
- **Record-id path encoding at the 128-byte cap** — `intervention/ids.rs` is its own module. `tc_471_intake.rs` asserts the grammar, the `p-`/`b-` disjointness over ids chosen to stress it, the cap arithmetic (173-byte basename, 178 with `.json` — unpadded base64url, so the trailing group is three symbols and not four), RFC 4648 §10's published vectors, and that every symbol the encoder can emit is in §5's table.
- **`resolveRawPath`'s escape refusals** — `tc_471_raw_evidence_resolution_refuses_every_escape` builds a real tree with real symlinks: one leaving the store is refused, one staying inside resolves to the same digest as its target, and absent / directory / lexical cases keep the oracle's sentences.
- **#407 declared** — `DIVERGENCE.md` §11.7, and honestly. `digest_file_sha256`'s symlink refusal is **unreachable** on this path, because resolution canonicalises before digesting. What actually diverges is the 256 MiB `MAX_DIGESTED_FILE_BYTES` ceiling and the non-regular-file refusal. The §11.7 text says that rather than parroting the ticket.
- **`InterventionIntakeError`'s code set** — a `Copy` `wire_enum!` with `as_str` / `from_code` / `ALL` and round-trip tests, per the repo idiom.

## Three types became two

`RawEvidenceClaim` (unchecked bare `String`s) is deleted. Two remain, and they mean different things: `RawEvidenceReference` is **proved** (only `raw_evidence_for` mints one, by resolving and digesting), `RecordedEvidenceReference` is **asserted** (what a record on disk says). `verify_raw_evidence_references` is the only way across, and `From<RawEvidenceReference>` goes one direction only.

`tc_471_evidence_reference_types.rs` reads the crate's own sources and fails if a third type carrying both a `path` and a `digest` appears, with anti-vacuity floors on both the file walk and the struct parse.

**I did this one.** quoin#472 (W5) carries the same obligation and does not need to.

## Duplication removed while porting

- `json_names(root)` — one directory listing, shared by measurements and interventions.
- `DiskMeasurement::resolve_raw_evidence` — one path resolution, shared by the text read and the digest, so the escape check exists once.
- `js_string` moved to `common/scalar.rs` — shared by the semantic pass and the agent-eval report reader.
- `common/document.rs` — one crossing between quoin-store's JSON model and serde's, through canonical bytes, so the two models cannot drift into two opinions about what a document is.

No second RFC 3339 grammar, no re-vendored schema, no `base64` crate (deny.toml forbids duplicate versions), no live node/pnpm/tsx spawn from any Rust test.

## Divergences declared

`quoin-measurement/DIVERGENCE.md` §11.1–§11.8: the dropped `blake3:` immutable-version alternative (quoin#409, narrowing); one RFC 3339 grammar for `observed_at`; ajv-vs-`jsonschema` finding text; the `RawEvidenceClaim` collapse; UTF-8 vs UTF-16 scenario-id ordering; the missing-`results` sentence where the oracle throws a bare `TypeError`; the three `digest_file_sha256` refusals; and the retained record's file name.

## Defect found and ticketed, not absorbed

**quoin#486** — `spec/evidence/interventions/quoin-270-cli-eval-sentinel-contract.json` has no `p-` prefix; it predates the namespacing in `intervention.ts:39-45`. Reads are unaffected (both implementations list the directory), but a re-publication by **either** implementation writes a second file beside it rather than over it, giving two records with the same `record_id`. The fix is a `git mv` in the store, not a change here. `tc_471_corpus.rs` asserts both halves so the rename fails that test and points at §11.8.

No pin weakened, no lint relaxed, no blanket allow added, no test deleted, no gate lowered.

Closes #471.


kreneskyp:
Rebased onto `f1c2bd9` (PR #487, quoin#472). The two waves are now one tree and the duplication is resolved toward one implementation each way.

## Dropped as already landed

- `tests/tc_471_evidence_reference_types.rs` — #487's `tests/tc_472_evidence_representations.rs` is the census for the three→two evidence-representation collapse and it is the one on `main`. One thing mine asserted that theirs did not: mine matched a retained-evidence pointer on `path` + `digest`, theirs on all four of `path`/`media_type`/`size_bytes`/`digest`. Two is the stronger net — a future type pointing at a retained file with only a path and a digest slips through a four-member match — and it still excludes the near miss the four-member spelling was written for (`RawEvidenceFile` has a size and a digest but no path). Folded into theirs with the reason and a `quoin#471` provenance line; no second census file.

## Collapsed

- **The JSON crossing → `src/json_bridge.rs`.** My `common/document.rs` is deleted and its two callers (`intervention/intake.rs`, `intervention/agent_eval/produce.rs`) import `crate::json_bridge`. `json_bridge` did everything mine did. Per your ruling, `quoin_core::ops::change_assurance::support::to_serde` is left alone.
- **`RecordedEvidenceReference::from(RawEvidenceReference)`.** My `From` impl is kept, and #487's `operational/github_release/records.rs` — which wrote the same four-field widening inline — now calls it.
- `lib.rs`'s module list, its retained-TypeScript table and both `mod.rs` files carry both waves' entries.

## What the two halves look like side by side

- **`intervention/semantics.rs` vs `operational/semantics.rs` — genuinely different.** Different record families, no shared cross-field rule. But they shared one helper by accident: `operational/semantics.rs` carried a private `rendered()` that is a second, narrower `String(x)` — numbers via `serde_json`'s `to_string` (`1` as `1.0`, `1e21` as `1e21`) and arrays as `[object Array]` where `String([1,2])` is `1,2`. It is `common::scalar::js_string` now, which is the ECMA-262 spelling. No retained record separates them (the schema constrains both pin members to strings), but a refused record's finding is the oracle's now rather than approximately it.
- **`intervention/validate.rs` vs `operational/validate.rs` — the same eight lines twice.** Both rendered a validator error as `` `${instancePath || "/"}: ${message}` `` and both collapsed findings to a sorted unique set, in two copies that nothing would have caught drifting. Extracted to `common/schema.rs` (`findings`, `sorted_unique`) and called by both. The compiled-validator `OnceLock` stays in each family's own module — the schema it compiles is the one thing about the pass that genuinely differs.
- **`intervention/intake.rs` vs `operational/intake.rs` — genuinely different** (pair envelope, store-wide lock, linkage check), and they already shared `InterventionIntakeError`. One duplication: `operational/intake.rs::validate_for_intake` hand-wrote three `MeasurementError` → `InterventionIntakeError` mappings that the exhaustive `From<MeasurementError>` impl already covers. They agreed on every code those three calls can return today and would have silently disagreed on the next one added, and they dropped the rendered sentence when a refusal carried no findings. Now `?`.

## Also

- **RFC 4648 §5 encoder in `intervention/ids.rs`:** nothing in the workspace has one. No `base64` crate dependency, no other hand-rolled encoder in any `rust/crates/*`. It stays, measured against §10's published vectors.
- **Module sizes:** `src/source.rs` reached **534** lines with both waves on it — over `tc_468_module_sizes.rs`'s 500-line soft ceiling, whose list must stay empty. Split on responsibility into `source/mod.rs` (235, the trait and `Clock`), `source/disk.rs` (197) and `source/memory.rs` (146); public paths unchanged. After the split the largest modules are `operational/record.rs` 498, `date_time.rs` 405, `agent_eval/produce.rs` 382. Nothing over 500, hard-ceiling allow-list still `&[]`.
- **`DIVERGENCE.md`:** both waves' sections kept. #487 owns §11, so wave 4's block is renumbered **§12** (§12.1–§12.7); my old §11.4 (the claim-type census) is dropped because #487's §11.7 records that collapse. §11.7-as-written — the `digest_file_sha256` / quoin#407 correction — is now **§12.6**, unchanged in substance.
- quoin#486 stays open. Nothing in `spec/evidence/` was moved.
- Zero `.ts` files touched; no pin weakened, no lint relaxed, no test deleted.

## Gates, re-run on the merged tree

| gate | result |
|---|---|
| `cargo fmt --all --check` | exit 0 |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | exit 0 |
| `cargo test --workspace --locked` | exit 0 — **967 passed, 0 failed** |
| `make rust-gate` | exit 0 |
| └ `rust-e2e` | 9 files, **47/47** |
| └ `rust-difftest` | **12 cases, 0 differences** |

kreneskyp:
## Review of the rebase — measured, not read

**The side-by-side comparison was worth asking for, and the best thing in this PR is a defect it found in code that merged an hour ago.**

`operational/semantics.rs` carried a private `rendered()` that claimed to be JavaScript's `String(x)` and was wrong on two of its seven arms:

```rust
Some(Value::Array(_)) => "[object Array]".to_owned(),   // String([1,2]) is "1,2"
Some(Value::Number(number)) => number.to_string(),      // serde_json's spelling, not ECMA-262's
```

`String([1,2])` is `"1,2"` and `String([])` is `""` — `[object Array]` is not a spelling JavaScript ever produces for an array. That shipped in #472 and I merged it. It was invisible in isolation and obvious the moment both halves sat on one tree, which is exactly the argument for doing this comparison at the merge rather than never. The replacement, `common::scalar::js_string`, recurses, renders `null`/`undefined` elements as empty per `Array.prototype.toString`, and routes numbers through `js_number_string` — the one number formatting in the crate — so the set key and the renderer cannot spell a number differently.

That is the right way to handle it: the defect got fixed, not absorbed, and not papered over.

**The other two duplications found by the same comparison** — both `validate.rs` files rendering schema errors and sorting/deduping findings identically (now `common/schema.rs`), and `operational/intake.rs` hand-writing three `MeasurementError` mappings the exhaustive `From` impl already covers *and dropping the message when a refusal carried no findings* (now `?`). The second of those is also a behaviour fix, not only a tidy-up.

**The merge broke a size ceiling and you split rather than widened.** `src/source.rs` hit **534** with both waves on it, over the 500 soft ceiling whose allow-list must stay empty. Split into `source/{mod,disk,memory}.rs` at 235/197/146 with public paths unchanged. That is the correct response — the alternative was an allow-list entry, and an entry there is a debt, not a dispensation. Largest modules now `operational/record.rs` 498, `date_time.rs` 405, `agent_eval/produce.rs` 382; list still `&[]`.

**Collisions resolved the way they should have been**
- `tc_471_evidence_reference_types.rs` dropped as already landed — and its genuinely stronger assertion (match on `path` + `digest` rather than all four members, which still excludes `RawEvidenceFile`) folded into #472's census with its provenance line. Keeping the better assertion without keeping a second file is the right call.
- `common/document.rs` deleted, callers onto `json_bridge.rs`. The `quoin-core` copy left alone per the ruling: the only shared home would be `quoin-store`, which has no `serde_json` dependency by design.
- Both waves' `lib.rs` and `mod.rs` entries kept.
- `DIVERGENCE.md` renumbered to §12 around #487's §11, with the #407 correction preserved in substance as §12.6.

**Your #407 correction stands and is the measured version.** `digest_file_sha256`'s symlink refusal is unreachable on the raw-evidence path because resolution canonicalises before digesting — that canonicalisation *is* the escape check. The real divergence is the 256 MiB ceiling and the non-regular-file refusal. The ticket's framing was wrong; yours replaces it.

**The RFC 4648 §5 encoder stays**, having checked nothing in the workspace already has one.

**#486 correctly untouched.** Nothing moved under `spec/evidence/` — retained bytes are frozen under NFR-025 and re-homing that record is its own change with its own review.

**Gates, re-run independently at `fdd86a8` in my own worktree**
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean.
- `cargo test --workspace --locked` — green.
- TypeScript touched: **zero files**.
- No pin, lint, test or gate weakened to make the merge pass.

Approved. Closes #471.

Issue #403, opened 2026-09-12
[Quoin#403] src/quire/validate.ts promises ajv's diagnostic order — Stage 1 cannot keep that promise


kreneskyp:
## The promise

`src/quire/validate.ts` documents its output lines as being "in the order ajv reported them". That is a comment describing a real property of the current implementation, and it is also a commitment the Rust port cannot honour.

## Why it cannot be kept

Measured during #378 across 118 documents:

- The `jsonschema` crate matches ajv's order on **110 of 118**.
- The two `jsonschema` releases under evaluation (0.18.3 and 0.56.0) disagree **with each other** on `exports-two-faults`.

So diagnostic order is not a stable property of the Rust validator across its own versions, let alone across engines. #378's crates sort deterministically on `(instance location, keyword)` instead.

## What this ticket wants

Correct the comment **before** the Stage 1 cutover, not after.

The distinction #378 established and this repo should adopt everywhere: **verdicts are contractual, diagnostic text and diagnostic order are not.** Verdict parity is exact (171/171 in #378). Order was never a contract; it was an accident of ajv that a comment turned into an apparent guarantee.

Correcting it afterwards reads as the port having lost a property. Correcting it now records the true state: the guarantee was never sound, and the replacement is a stronger one — a deterministic sort that does not vary with the validator's version.

Related: #378 (PR #402), where the measurement was made; #379 (PR #398), Stage 1.

Issue #387, opened 2026-09-12
[Quoin#387] [Stage 6] quoin-measurement-graph: the governed graph portfolio


kreneskyp:
Port `src/measurement/graph-adapters.ts`, `graph-portfolio.ts`, `graph-portfolio-load.ts` (1,812 lines, 3 files) to `rust/quoin-measurement-graph`.

**The one crate boundary in Stage 6 the code actually supports.** Single entry point (`graph-portfolio-load.ts:21 buildGovernedGraphPortfolio`), no inbound edges except the facade `index.ts`, and it is the only part of the measurement subsystem that reaches `../quire/index.js` (`graph-adapters.ts:8 validateAssurance`) and `../graph-analysis/index.js` (`graph-portfolio-load.ts:8`). Splitting it isolates those two dependencies from the rest of measurement, which is what a crate boundary is for.

**Depends on #385** (`quoin-graph-analysis`) — land that first — and on #380 for `canonicalJson` (`graph-adapters.ts:7`, `graph-portfolio.ts:6`).

## Acceptance
- governed-graph portfolio output byte-identical to TS on a golden corpus
- `graph-portfolio.ts:326 canonicalGraphPortfolioJson` delegates to `quoin-store`'s canonical function — it is a thin wrapper today and must not become a second implementation
- FR-066/FR-067 adapter behaviour asserted by Rust tests
- clippy clean at `-D warnings`

---
Parent: #373. Depends on the workspace from #375 for wiring, but the crate itself is written and tested standalone — do not block on it.

Standing rules: `/rust-review` is the style guide, a repo `rust-style` doc outranks it. `#![forbid(unsafe_code)]`, one `thiserror` error enum per crate boundary with stable codes, newtypes over bare `String`/`Uuid` for ids. The existing TypeScript is the oracle: capture expected fixtures from it once, never leave TS as a runtime oracle. **No TypeScript is deleted in this ticket** — cutover and deletion are separate tickets per FR-018.

kreneskyp:
Delivery reconciliation: #387 is complete. quoin-measurement-graph was delivered through 196cfcb, d2e556a, d54cce6, and d2ba6eb; its Rust golden/criterion suites cover governed-graph loading, canonical output, and FR-066/FR-067 behavior. The TypeScript governed graph portfolio was deleted only after the Rust cutover in dc6bcec / eb0ca08. Closing this stale implementation ticket; later promotion/release work belongs to #520/#521, not this completed Stage 6 port.

PR #489, opened 2026-09-13
feat(quire): validate_assurance over a vendored assurance-v1 schema

### LINKED ISSUE #474: [Stage 6] quoin-quire must expose validate_assurance over the vendored assurance-v1 schema
Wave 7 of the Stage 6 measurement port, concurrent with W6. `quoin-quire` gains `validate_assurance` over a vendored `assurance-v1.schema.json` — the gap that blocks the graph adapters.

Needs `quoin-jsonschema` (#470) and the vendored schema.

Gate: the vendored schema is byte-equal to `src/quire/schemas/assurance-v1.schema.json`; verdict parity with `src/quire/validate.ts:99` on a corpus; #403's diagnostic-ordering promise explicitly not made.

Trace: FR-099. Blocks #387.

Parent: #373

kreneskyp:
Wave 7 of Stage 6. `quoin-quire` gains `schema::validate_assurance`, the Rust
successor to `validateAssurance` (`src/quire/validate.ts:99`) — the gap that
blocks the graph adapters (#387, W8/W9).

**Not a cutover.** Zero TypeScript deleted, zero `.ts` files touched.
`src/quire/validate.ts` remains the oracle and was not improved on the way out.

## What was added

968 lines of Rust added, 4 removed, across seven files:

| file | lines | what |
| --- | ---: | --- |
| `src/schema.rs` | 191 | the vendored schema, `ValidAssuranceDocument`, `validate_assurance` |
| `src/error.rs` | +52 | `QQ-1063 AssuranceContract`, `QQ-1064 VendoredSchemaInvalid` |
| `src/lib.rs` | +9/-4 | the re-export and the capability table |
| `oracle/capture-assurance-bases.rs` | 139 | one-time: the corpus base documents |
| `tests/tc_474_assurance_schema.rs` | 197 | provenance and the format policy |
| `tests/tc_474_assurance_parity.rs` | 250 | verdict parity against the oracle |
| `tests/tc_474_module_sizes.rs` | 130 | the crate's module-size census |

Plus `schemas/assurance-v1.schema.json` (7,712 bytes, vendored),
`oracle/capture-assurance-verdicts.mjs` (the one-time ajv capture), two goldens
and `DIVERGENCE.md`.

## Which constructor, and why

`SchemaValidator::compile_vendored` — which registers no format check and
asserts none.

`src/quire/validate.ts:99` is `new Ajv2020({ allErrors: true, strict: false })`
and registers **nothing**. `assurance-v1` declares one `format` (`uuid`, on
`$defs.artifact.uuid`), and under that ajv it is an annotation: ajv logged
`unknown format "uuid" ignored in schema at path "#/properties/uuid"` while the
capture ran, and accepted `"not-a-uuid"`. That is the measurement the choice
rests on, not the fact that the schema has a `format`.

`compile_with_formats` — which the two *measurement* schemas need, because their
ajv instances each register `date-time` (`quoin-jsonschema/DIVERGENCE.md` §4) —
would have refused documents the oracle accepts. `tc_474_012` asserts it from
both sides: the same document is accepted by `validate_assurance` and refused by
a build of the same vendored schema with a `uuid` check registered, so the
constructor cannot quietly invert without a red test.

`quoin-jsonschema` is the only validator involved. No `jsonschema` dependency
was added to `quoin-quire`; that pin stays in `quoin-jsonschema`'s manifest.

## The corpus

**217 documents**, of which the oracle accepts 28 and refuses 189. It is
`base documents × a fixed named mutation ladder`, following
`quoin-jsonschema/tests/tc_470_parity.rs` — not hand-written fixtures, not
sampled.

The two bases are real exports produced by the linked engine
(`quire_rs::build_assurance_export`, rev `85dfe9d5`) over the two committed
fixture corpora: `tests/fixtures/repo` under its own module, where every
collection is populated, and under `tests/fixtures/module-no-traceability`,
where `artifacts`, `obligations`, `relation_kinds` and `relations` come back
empty. There is **no retained `assurance-v1` document in this repository** to
read instead, and the engine is the only thing that can write one — so the bases
are captured once and committed with their provenance. Not a self-fixture: the
thing under test is the schema check, which had no hand in producing them.

Each base is then mutated once per named rung: every top-level key removed in
turn, every key of the first element of every collection removed and nulled in
turn, an unknown key added at each level, the two `const` tags broken, the
`revision` and `digest` patterns broken, and every arm of the only lookahead in
the document (`locator.path`'s `^(?!/)(?!.*(?:^|/)\.\.(?:/|$)).+$`) — including
the three arms that must be **accepted** (`a/..b.md`, `a/b...md`, `a/./b.md`),
so a regex that refused everything could not pass as agreement. Both bases carry
both verdicts, asserted per base. `CORPUS_FLOOR = 180` is the anti-vacuity floor;
`tc_474_022` additionally names the twelve load-bearing rungs so the corpus
cannot shrink quietly.

The ajv capture is a one-time `.mjs` that **imports** `validateAssurance` rather
than rebuilding its ajv instance, and writes a frozen golden recording ajv
8.20.0, node v22.15.0, the engine pin and the quoin revision. No Rust test
spawns node, pnpm or tsx (FR-101-AC-5).

## Divergences

**`DECLARED_DIVERGENCES` is empty**, and the test makes that a measured claim:
an undeclared disagreement fails, and a declared one that stopped disagreeing
fails too. All 217 verdicts agree.

`DIVERGENCE.md` records the rest:

1. Verdicts are contractual. **Error text, error count and error order are not** —
   #403's ordering promise is explicitly not made, and no test asserts any of the
   three.
2. `format` is annotated, never asserted, and why.
3. `ValidAssuranceDocument` is a local newtype rather than a reuse of
   `quoin_jsonschema::ValidDocument`: that one is indexed by `VendoredSchema`,
   the closed enum of the two *measurement* documents, whose `compile()`
   hardcodes a `date-time` check. Widening it would file a quire output contract
   under the measurement namespace and hand it exactly the wrong format policy.
   The validator is shared; only the ~15-line proof token is local.
4. No declared verdict divergence.
5. `validate_assurance` is the weaker of the two readers; a caller with premises
   wants `assurance::read`.

## Defects found

None absorbed, none covered up. Two pre-existing modules sit above the 500-line
soft ceiling (`error.rs`, `properties.rs`); the new census names them by name
rather than raising the ceiling, and `error.rs` grew by this ticket's two
variants. Nothing was pinned down, lint-allowed or deleted to make a gate pass.

`Error` carries `Vec<String>` violations rather than `Vec<SchemaError>` because
`Error` derives `Clone, PartialEq, Eq` and `SchemaError` is not `Eq` — carrying
it would have meant weakening the derives.

## Gates

Run from `rust/` with a worktree-local `--target-dir`, all exit 0:

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked` — 0 failed
- `make rust-gate` — fmt, clippy, cargo deny, tests, the end-to-end caller and
  the differential harness (12 cases, 0 differences)

Nothing in `quoin-measurement` was touched.

Closes #474.


kreneskyp:
## Review — approved, merging

Gates re-run independently at `ddd69aa` on a clean worktree, not taken from the agent's report:

- `CLIPPY_EXIT=0`, `TEST_EXIT=0`, 115 test binaries green
- zero `.ts` files in the diff (14 files changed)
- vendored `assurance-v1.schema.json` is byte-identical to `src/quire/schemas/assurance-v1.schema.json` (`cmp` clean), so the vendoring is a copy and not a reinterpretation

Three things make this credible rather than merely green:

**`compile_vendored` was chosen by measurement.** `validate.ts:99` registers no format check, and ajv itself logged `unknown format "uuid" ignored` during the capture. `tc_474_012` asserts the choice from both sides, so the decision cannot invert silently later.

**The empty `DECLARED_DIVERGENCES` across 217 agreeing verdicts is the finding.** It is believable because the corpus is not refuse-shaped: 28 accepted / 189 refused over two engine-produced bases × a named mutation ladder, both bases carrying both verdicts, `CORPUS_FLOOR = 180`. Three of the ten `locator.path` lookahead arms must be *accepted*, so a refuse-everything regex could not masquerade as parity.

**The new census records debt, it does not raise a ceiling.** `ALLOWED_ABOVE_HARD_CEILING` is empty. `OVER_SOFT_CEILING = ["error.rs", "properties.rs"]` names two modules that predate this ticket, and `tc_474_031` asserts *exact set equality* — a third crossing fails, and so does splitting one out without recording it.

`ValidAssuranceDocument` follows parse-don't-validate (`RawFileSha256Digest::parse_stored` is the in-repo model). No `jsonschema` dependency added to `quoin-quire`. #403's diagnostic-ordering promise is explicitly *not* made, which is the right call under 'verdicts are contractual, error text is not'.

PR #490, opened 2026-09-13
feat(rust): port the measurement report and portfolio views (wave 6)

### LINKED ISSUE #473: [Stage 6] quoin-measurement reporting: the measurement report, comparisons, series and the portfolio view
Wave 6 of the Stage 6 measurement port. Ports `report.ts` (384) and `portfolio.ts` (371) — 755 lines.

Needs waves 1, 2, 4, 5 and `quoin_store::canonical_json`.

Gate: `renderMeasurementReport`, `renderMeasurementReportJson`, `renderPortfolioReport`, `renderPortfolioReportJson`, `comparisonFor` and `seriesFor` byte-identical on a committed multi-repo fixture.

Trace: FR-100-AC-4.

Parent: #373

kreneskyp:
Wave 6 of Stage 6. Ports `src/measurement/report.ts` (384 lines) and
`src/measurement/portfolio.ts` (371 lines) into the existing
`quoin-measurement` crate as `src/report/` and `src/portfolio/`.

**Not a cutover.** No TypeScript was deleted and no `.ts` file was touched. The
retained tree stays the oracle until quoin#479.

## What landed

**2,966 lines of Rust added, 18 removed** across 13 new modules and 8 edited
ones, plus a 4,165-line total diff including the fixture and the capture script.

| retained TypeScript | here | lines |
| --- | --- | --- |
| `report.ts` | `report/{build,comparison,render,render_json,series,wire}.rs` | 1,458 |
| `portfolio.ts` | `portfolio/{build,location,render,render_json,types}.rs` | 1,056 |

Largest new module is `report/wire.rs` at 365 lines. Nothing is over the 500-line
soft ceiling, and `tc_468_module_sizes`'s over-soft allow-list is still empty.

## The gate

All six functions are **byte-identical** to the retained TypeScript on a
committed multi-repo fixture (FR-100-AC-4), asserted by `tc_473_reporting`:

- `renderMeasurementReport` and `renderMeasurementReportJson` — 5 repository
  cases
- `comparisonFor` — 4 cases (baseline found, baseline matched by prefix, no
  baseline, no current collection)
- `seriesFor` — 4 cases
- `renderPortfolioReport` and `renderPortfolioReportJson` — one walk over 8
  locations

### What the fixture actually is

`tests/fixtures/portfolio-tree/` — **21 committed documents** across seven
locations, authored for this wave rather than copied from any Rust output:

- `alpha` — 8 assurance markdown documents (2 profiles, 6 plans, one retired of
  each) and 3 measurement collections, exercising a v1 and a v2 collection, a
  present and an **absent** `dimensions`, a `population` with unmodelled
  `exclusions`/`namedMisses`, a `not_computed` with a reason and one without, a
  metric whose stored definition has drifted from its active plan, an authored
  plan no collection computes, a retired plan, and the `1e+21` number spelling
- `bravo` — a plan and an empty store
- `charlie` — a profile and no plans at all
- `delta` — 120 days stale
- `golf` — a collection whose `timestamp` no date grammar accepts
- `echo-not-a-directory` — a regular file
- `zulu-does-not-exist` — passed to the walk but never created

Anti-vacuity floors are asserted on every case list (5 / 4 / 4 / 8 locations),
and the capture's `produced_from_revision` must be 40 characters.

### Retained artifact vs captured golden

**None of the six functions has a retained rendered artifact to reproduce** —
quoin retains measurement collections and evidence records, not rendered
reports — so all six are proven against the captured golden
`tests/fixtures/portfolio-oracle.json`, taken once by
`oracle/capture-portfolio-oracle.mjs` at revision `9e7c33e` on node v22.15.0 and
frozen with that provenance. No test spawns node, pnpm or tsx (FR-101-AC-5).

Two **retained artifacts are reproduced byte-for-byte inside the fixture**: the
intervention record `quoin-270-cli-eval-sentinel-contract.json` and the
operational pair `c1b30a18….json`, copied unchanged from `spec/evidence/`. Both
are read, projected and rendered into `alpha`'s report, so the intervention and
operational sections of the gate run over real retained bytes.

## Defects found, fixed rather than absorbed

**1. Unmodelled `population` members were dropped.** `DIVERGENCE.md` §3.3 said
they were and added "Nothing in the corpus carries one." That was false: **25 of
14,644** observations under `spec/evidence/measurements/` carry `exclusions` and
`namedMisses`, and `renderMeasurementReportJson` writes the population straight
back out, so the drop was a real byte divergence. `MeasurementPopulation` now
keeps every unmodelled member verbatim and flattens it back. §3.3 is amended to
say so, with the count.

**2. An absent `dimensions` was flattened to an empty map.** **305 of 14,644**
observations state no `dimensions` at all, and `{}` is not the same bytes as
absent. The new `Dimensions` newtype models the distinction — `ABSENT` or
`stated(entries)` — rather than a `bool` or an `Option` threaded through call
sites.

A census over all 14,644 retained observations settles what is left: no
unmodelled observation-level or collection-level member, `dimensions` is never a
stated `{}`, `population` is never a non-object, all four modelled population
members always carry their modelled type, and all 23,214 `dimensions` and
`population` objects are already stored in sorted-name order.

**3. A gate matched a substring where it meant a member name.**
`tc_472_evidence_representations` asked whether a struct body contained the text
`"digest:"`, which every `config_digest:` member also contains, so six
measurement-collection references read as retained-evidence pointers. It now
matches parsed member names. This is a **stricter** net, not a looser one: a
type declaring `digest` still matches and one declaring only `config_digest` no
longer pretends to. No allow-list entry was added and nothing was deleted.

## Divergences — `DIVERGENCE.md` §13

- **§13.1** the fixture compares with one path substitution (`@@TREE@@` for the
  tree's absolute root), because `root` and `path` are absolute on both sides.
  Neither side canonicalises, so a symlinked checkout still agrees.
- **§13.2** a caught reader error renders Rust's message, not node's. The three
  messages quoin constructs itself are reproduced verbatim and are under the
  byte comparison; a message raised by node's `fs` or `JSON.parse` is not, and
  that one string diverges.
- **§13.3** staleness reads instants with `Rfc3339DateTime` rather than
  `Date.parse`, per the Stage 6 plan §5 ruling. Every corpus timestamp parses
  under both.
- **§13.4** the two fidelity defects above, with their counts.
- **§13.5** `dimensions` re-serialise in sorted-name order rather than insertion
  order — invisible in canonical JSON, in the rendered suffix, and in the
  corpus, all of which are already sorted.

## Gates

Run from `rust/` with a target directory inside the worktree, unpiped:

| gate | exit |
| --- | --- |
| `cargo fmt --all --check` | 0 |
| `cargo clippy --workspace --all-targets --locked -- -D warnings` | 0 |
| `cargo test --workspace --locked` | 0 |
| `make rust-gate` | 0 |

Worktree and both target directories removed.

Closes #473.


kreneskyp:
## Review — approved, merging

Gates re-run independently at `b2b2770` on a clean worktree, not taken from the agent's report: `CLIPPY_EXIT=0`, `TEST_EXIT=0`, 115 test binaries, zero `.ts` files in the 46-file diff.

### The three defect counts check out exactly

I re-measured them myself with `jq` over `spec/evidence/measurements/` rather than accepting the numbers: **14,644** observations, **25** carrying `exclusions`/`namedMisses`, **305** with no `dimensions` key. All three match.

Both were real byte divergences on re-serialisation, both shipped in W1, and both are fixed here rather than absorbed — with §3.3 *amended* to withdraw its false 'nothing in the corpus carries one' rather than left standing next to a contradicting §13.4. The `Dimensions` newtype modelling absent-vs-`{}` is parse-don't-validate applied to a distinction that `Option<BTreeMap>` alone would have lost on the way out.

### The change to the merged `tc_472` census is stricter, not looser

It moved from the substring `"digest:"` to an exact parsed member name, so `config_digest` no longer masquerades as a retained-evidence pointer. A type declaring `digest` still matches. No allow-list entry, no deleted test, no lowered gate — which is the required shape when a better rule surfaces a defect.

### §13.3's narrowing is provably unreachable

The divergence claims every retained `timestamp` parses as RFC 3339, so the `Date.parse` widening cannot bite. I checked: **0** of the corpus timestamps fail an RFC 3339 match. The refusal path is still under the byte comparison because the fixture's `golf` repository carries `"timestamp": "the-first-of-never"`, which both sides reject.

### §13.2 is the part worth naming

The agent declared a divergence it could **not** put under the byte comparison — a portfolio run over a corrupt or unreadable collection file renders this crate's error text, not node's `ENOENT`/`JSON.parse` prose — and stated why reconciling it is forbidden (a second error vocabulary inside the crate, which §2.3 and Stage 6 plan §5 both bar). Declaring the unreached case beats widening the fixture until it looks covered.

### Structure

13 new modules across `src/report/` and `src/portfolio/`, largest 365 lines, everything under the 500 soft ceiling and nothing added to an allow-list. No test spawns node — the oracle is a frozen committed capture (FR-101-AC-5), with anti-vacuity floors on every case list.

Issue #476, opened 2026-09-13
[Stage 6] quoin-measurement-graph projection: the FR-067 governed graph portfolio


kreneskyp:
Wave 9 of the Stage 6 measurement port, concurrent with W8. Ports `graph-portfolio.ts` (881) into `quoin-measurement-graph`.

Needs `quoin_measurement::portfolio` (W6) and `quoin_store::canonical_json`.

Gate: governed-graph portfolio output byte-identical on a golden corpus; `canonicalGraphPortfolioJson` (`graph-portfolio.ts:326-329`) delegates to `quoin_store::canonical_json` and a static check proves there is no second implementation.

Trace: FR-100-AC-4, FR-100-AC-8.

Parent: #373

kreneskyp:
Two corrections from #475, which landed the shared crate (PR #491, merged `73f11a8`):

1. **The '107 zod declarations' figure in this ticket and #475 matches no measurement of either file.** Measured on `graph-adapters.ts`: 112 `z.*(` constructor calls, 29 `.strict()` objects, 24 named `*Schema` consts. Do not treat the figure as ground truth — measure the file and assert the measured number from both sides, which is what #475's census now does.

2. **The `Buffer.from(x, "base64")` decode call site is in *this* ticket's file**, `graph-portfolio.ts:786`, not in `graph-adapters.ts` — which only encodes, at `:564`. #475 therefore built the strict decoder as the encoder's inverse and declared the quoin#465 leniency divergence at its own boundary. This wave must call `crate::base64::decode` rather than write a second decoder, and extend the existing #465 divergence entry rather than adding a parallel one.

Also already landed by #475 and not to be duplicated: `crate::canonical` (`pretty_text`, `pretty_text_trimmed`, `compact_bytes`), which delegates to `quoin_store::canonical_json`/`canonical_bytes` through the one `quoin_measurement::json_bridge` — **this already satisfies this ticket's FR-100-AC-8 criterion**; `crate::error::{GraphAdapterError, GraphAdapterErrorCode}`; and the crate's single census, `tests/tc_475_module_sizes.rs`.

Issue #465, opened 2026-09-13
[Quoin#465] Retained graph evidence may carry base64 payloads Node accepted and a strict decoder refuses


kreneskyp:
`Buffer.from(x, "base64")` ignores invalid characters; a strict Rust decoder refuses them. `src/measurement/graph-portfolio.ts:785` decodes base64 before digesting.

Population is **unmeasured**. Stage 6 W8 measures it before writing the decoder; an empty population is a valid result worth recording.

Surfaced by the port making a permissive rule strict. Not a port blocker.

PR #491, opened 2026-09-13
feat(rust): digest_bytes_sha256 and the graph-adapters port (#484, #475)

### LINKED ISSUE #475: [Stage 6] quoin-measurement-graph adapters: the FR-066 versioned adapters for retained graph evidence
Wave 8 of the Stage 6 measurement port, concurrent with W9. Ports `graph-adapters.ts` (763) into the new `quoin-measurement-graph` crate.

Needs `quoin_quire::validate_assurance` (W7), `quoin_store::canonical_json`, and a new `quoin_store::digest_bytes_sha256`.

Gate: 107 zod declarations become `deny_unknown_fields` serde types; FR-066 adapter behaviour asserted; the `Buffer.from(x, "base64")` leniency is declared as a divergence — Node ignores invalid characters, so a strict Rust decoder newly refuses input Node accepted (#465).

Trace: FR-100-AC-4, FR-100-CON-4.

Parent: #373

### LINKED ISSUE #484: quoin-store: expose digest_bytes_sha256 for in-memory raw evidence
Two Stage 6 waves need a bytes-wise sha256 from `quoin-store` and there is no public one.

`sha256_hex` is private and `digest_raw_bytes` is blake3, so an in-memory source has no way to account for raw evidence without minting a second sha256 — which FR-100-CON-4 forbids.

Consequences today, both handled by refusing rather than faking:
- **#468 (landed):** `MemoryMeasurement::raw_evidence_file` refuses with `QM-RAW-EVIDENCE-UNAVAILABLE`, asserted by `tc_468_a_memory_source_refuses_to_digest_rather_than_minting_a_second_sha256`. Raw-evidence accounting is disk-only until this exists.
- **#475 (W8, graph adapters):** named as a dependency in the wave plan.

Fix: make the existing sha256 public as `digest_bytes_sha256` returning `RawFileSha256Digest`, in the one place `digest.rs` already owns hashing. Do not add a second call site — `digest.rs`'s own doc comment says the module's guarantee is gone if a third hashing call site appears.

Then remove the `MemoryMeasurement` refusal and flip that test to assert the digest.

Parent: #373

kreneskyp:
Wave 8 of the Stage 6 measurement burn-down: two tickets, two commits.

## `9120e9b` — `quoin-store::digest_bytes_sha256` (#484)

`quoin-store`'s digest module states its own guarantee in its doc comment:
one hashing call site, so a raw-file digest and a stored digest cannot
drift. Commit 1 exposes a bytes-wise digest that routes through that same
call site rather than adding a second one, and the in-memory measurement
source stops refusing (`QM-RAW-EVIDENCE-UNAVAILABLE`) to account for bytes
it is holding in its hand. The test that asserted the refusal now asserts
the digest, and is renamed to say so.

No `serde_json` was added to `quoin-store`; it still has none, by design.

## `196cfcb` — `graph-adapters.ts` to `quoin-measurement-graph` (#475)

`src/measurement/graph-adapters.ts` (763 lines) becomes a new workspace
crate of sixteen modules, none over 226 lines. `tests/tc_475_module_sizes.rs`
is its one census; `ALLOWED_ABOVE_HARD_CEILING` is empty, and so is the
soft-ceiling list.

Nothing is reimplemented: canonical bytes are `quoin-store`'s, the sha256 is
the `digest_bytes_sha256` of commit 1, the `serde_json`/`JsonValue` crossing
is `quoin_measurement::json_bridge`, the instant grammar is
`Rfc3339DateTime`, `assurance-v1` is checked by
`quoin_quire::validate_assurance`, and the transcribed collection is minted
by `quoin_measurement::validate::measurement_collection`.

### The corpus

`tests/tc_475_parity.rs` runs **89 cases — 45 accepted, 44 refused** —
captured once from the TypeScript by
`oracle/capture-graph-adapter-verdicts.mjs`, across `selectGraphAdapter`,
`graphQualityObservationId`, `adaptQuireAssurance`,
`adaptGraphQualityObservation`, `base64Encode` and `base64Decode`. The
corpus exercises all eight retained refusal codes, and a test asserts both
verdicts are present for every entry point — a refuse-everything corpus
agrees perfectly and proves nothing. Modelled on
`quoin-jsonschema/tests/tc_470_parity.rs`.

**The capture script is one-time. The W12 cutover commit deletes
`oracle/capture-graph-adapter-verdicts.mjs` together with the TypeScript
module it captures.** The TypeScript is never a runtime oracle (FR-101 AC-5).

### Declared divergences — 15 entries in two families

1. **quoin#465 — `Buffer.from(x, "base64")` leniency.** Node silently
   discards characters it cannot read; this decoder refuses them. Twelve
   malformed spellings Node accepted are newly refused.
2. **`Date.parse`'s implementation-defined heuristic.** V8 accepts
   `2026-01-02`, `2026/01/02 10:00` and `Jan 1 2026`; the workspace's one
   RFC 3339 grammar does not.

Every declared divergence must fire — the test asserts set equality, not
containment, so a divergence that stops firing fails.

A sixteenth difference is **not** declared but **proved unreachable**:
`graphQualityObservationId` sorts member names by Unicode code point, which
agrees with neither of `quoin-store`'s two canonical orderings. All member
names in an admissible record are schema-fixed ASCII, where the three orders
coincide; `tc_475_054` proves both that the orderings really differ and that
no admissible record can reach the difference. A declaration that cannot
fire is a licence, not a gate.

### The zod census

The ticket says 107 zod declarations. No measurement of the file yields
107: it has **146 zod constructor calls, 29 `.strict()` objects, 12 closed
sets, 13 literals, 7 optionals, 1 nullable, 24 named `*Schema` consts**.
`tests/tc_475_schema_census.rs` asserts the measured numbers from both
sides, so a silently dropped type fails rather than passing unexercised.

### Findings

- `Buffer.from(x, "base64")` decoding is **not** in `graph-adapters.ts` —
  it is at `src/measurement/graph-portfolio.ts:786`. This crate implements
  the strict decoder as the encoder's inverse and declares the divergence
  at its own boundary.
- `graphQualityObservationId` uses a third key ordering (see above).

No TypeScript is touched.

## Gates

`cargo clippy --workspace --all-targets -- -D warnings`, `cargo test
--workspace`, `cargo deny` and `cargo fmt --all --check` are all green on
toolchain 1.98.1.

Closes #484
Closes #475


kreneskyp:
## Review — approved, merging

Gates re-run independently at `196cfcb` on a clean worktree: `CLIPPY_EXIT=0`, `TEST_EXIT=0`, 122 test binaries, zero `.ts` files in the 35-file diff. Two commits in the right order — #484 then #475.

### #484 does not cost `quoin-store` its guarantees

`quoin-store/Cargo.toml` is **unchanged** — no `serde_json`, no private `sha2` call site. `digest_bytes_sha256` routes through the same private `sha256_hex` as every other sha256 entry point, so the module header's one-call-site-per-algorithm promise is intact, and the doc says explicitly why the file guards are absent (there is no file) and that a caller holding a *path* must still use `digest_file_sha256`.

`tc_484` asserts **published NIST vectors** rather than agreement with `digest_file_sha256`, and states the reason: both route through `sha256_hex`, so a self-comparison would agree with itself if `sha256_hex` were wrong. I checked both vectors against `sha256sum` — `abc` and empty, both correct. The `assert_ne!` against `digest_raw_bytes` keeps the domain non-decorative.

### `tc_475_054` is a proof, not a claim

The code-point vs code-unit ordering difference in `graphQualityObservationId` is **not** declared as a divergence — it is proved unreachable. That is the harder and better option, and the test earns it three ways: an anti-vacuity floor (>20 distinct member names across accepted records), every observed name asserted ASCII, and then a demonstration that the two orders *actually differ* on astral input. Without that last assertion the ASCII check would be vacuous. This is the right shape.

### `canonical.rs` is the FR-100-AC-8 answer, already

It delegates to `quoin_store::canonical_json` / `canonical_bytes` through the one `quoin_measurement::json_bridge`, adds no fourth writer, and says so in a doc comment that names all three existing writers. No second canonicalizer, no second bridge, no second error vocabulary.

### The ticket's number was wrong and the agent was right to correct it

My #475 text said '107 zod declarations'. It matches no measurement of the file — I independently get 112 `z.*(` calls, 29 `.strict()` objects and 24 named `*Schema` consts. The census now asserts the **measured** counts from both sides, which is worth more than the figure I guessed. Same correction applies to #476.

### Corpus

89 cases captured once from the TypeScript and frozen, **45 accepted / 44 refused**, `CORPUS_FLOOR = 80` with a `VERDICT_FLOOR = 20` on each verdict separately — so agreement on a one-sided corpus cannot masquerade as parity. All 8 retained refusal codes exercised. 15 declared divergences in two families (quoin#465 base64 leniency, `Date.parse`'s V8 heuristic), asserted by **set equality** so an unfired entry fails as loudly as an undeclared difference.

19 modules, largest `scalars.rs` at 338 lines, both census allow-lists empty.

### One cross-wave note, already relayed

The `Buffer.from(x, "base64")` **decode** call site is `graph-portfolio.ts:786` — W9's file, not this one. This PR implements the strict decoder as the encoder's inverse and declares #465 at its own boundary, which is correct. I have told the running W9 agent to call `crate::base64::decode` rather than write a second decoder, and not to re-declare #465.

Issue #477, opened 2026-09-13
[Stage 6] quoin-measurement-graph loader: the filesystem boundary for the governed graph portfolio


kreneskyp:
Wave 10 of the Stage 6 measurement port. Ports `graph-portfolio-load.ts` (168) — the filesystem boundary for the governed graph portfolio.

**Blocked on #385.** Needs `quoin_graph_analysis::{load_graph_analysis_input, analyze_fan_out, analyze_churn, analyze_change_impact}`.

Gate: `buildGovernedGraphPortfolio` end-to-end byte-identical.

Trace: FR-100-AC-4.

Parent: #373

kreneskyp:
Follow-up found by wave 11 (quoin#478): `make lint` went red across the whole repository from PR #494 onward.

`tests/fixtures/graph-loader-tree/delta/graph/export.json` is a deliberately truncated document — a single opening brace — so that `tc_477_011_graph_refusal_verdicts` can measure the verdict the loader gives an export a schema validator rejects. That is the right fixture. But `prettier --check .` cannot parse it, so the run **errored** rather than reporting a diff, and every file after it went unchecked. Twenty-seven sibling goldens and fixtures from #469, #470, #476 and #477 were reported unformatted by the same run.

The gate did not catch it at merge because the repo's test workflows are `workflow_dispatch:` only, and I reviewed #494 with the cargo gates rather than `make lint`. My miss, recorded here rather than absorbed.

Fixed in PR #497 on the rule `.prettierignore` already stated for earlier goldens: `tests/goldens/`, `tests/fixtures/` and vendored `schemas/` are evidence and are ignored; `oracle/*.mjs` and `DIVERGENCE.md` are source and prose and stay formatted. I verified no golden byte was rewritten — the PR touches no file under any `tests/goldens/` directory — and that every `.mjs` change is line-wrapping only, with no semantic token added or removed, so no captured oracle was invalidated.

PR #492, opened 2026-09-13
feat(rust): port graph analysis to quoin-graph-analysis (#385)

### LINKED ISSUE #385: [Quoin#385] [Stage 5] quoin-graph-analysis: fan-out, churn, change-impact
Port `src/graph-analysis/` (1,373 lines, 5 files) to a Rust crate `rust/quoin-graph-analysis`.

Cycle-free. The only Stage 5 directory with a consumer outside `src/commands/` — `src/measurement/graph-portfolio-load.ts:3` imports `analyzeChangeImpact`, `analyzeChurn`, `analyzeFanOut`, `loadGraphAnalysisInput`. **That is the independent-shipping argument for it being its own crate**, and it also means Stage 6's `measurement-graph` depends on this one. Land this before that.

## One shared decision, recorded in both leaf tickets

This crate needs exactly four types from auditor — `AuditReport`, `Finding`, `UnevaluatedCheck`, `Severity` (`src/auditor/audit.ts:38-144`) — and nothing else. Three other places want the same four: `src/assurance/graph.ts:14`, `src/graph-analysis/analysis.ts:3`, `src/graph-analysis/input.ts:5`, `src/change-assurance/types.ts:1`.

Once #376 lands, depending directly on `quoin-auditor` is *legal* — no cycle. It is just heavy: a leaf view crate would pull a 1,800-line engine crate to name four structs. **Recommend a small `quoin-finding-types` crate** holding those four. Cheap, and it keeps the view crates genuinely leaf. Not a blocker either way — decide it once, in whichever of these two tickets is picked up first, and make the other follow.

## Seams
One filesystem seam, and it is the whole IO surface: `src/graph-analysis/load.ts:3` imports `existsSync, readFileSync`, used at `:90` (existence probe), `:102` (`JSON.parse(readFileSync(...))`), `:175` (raw read). One trait: `GraphInputReader`. `analysis.ts`, `input.ts` and `render.ts` are pure functions.

No subprocess. No network. No `process.env` or `process.cwd`.

Value-level outbound edges to watch when wiring: `load.ts:7` (`bindingsPath`, `STORE_SCHEMA_VERSION` from evidence), `load.ts:9` (`parseAssurance` from quire), `render.ts:3` (`canonicalJson` from evidence). `canonicalJson` and `STORE_SCHEMA_VERSION` come from `quoin-store` (#380), so this crate depends on it.

## Generic vs quoin-specific
The algorithms are generic — `analysis.ts:154 analyzeFanOut`, `:197 analyzeChurn`, `:263 analyzeChangeImpact` are fan-out, churn and reverse-reachability over a graph; only their input types are quire-shaped. So are the canonicalizers `input.ts:178 canonicalizeAcceptedAssurancePremises` and `:202 canonicalizeAuditReport`.

Do not let those canonicalizers grow a second serialization implementation: they must call `quoin-store`'s single digest/canonical-JSON function (#380), not their own. One canonical form in the program, not two that agree by coincidence.

Naming note: `analysis.ts:523 selectedCorpusEdges` and `:529 AssuranceCorpusRelation` use "corpus" to mean *spec-corpus graph edges*, not evidence-corpus accounting. Not the corpus-accounting capability EA owns; do not route it there.

## Acceptance
- fan-out, churn and change-impact results identical to TS on a golden corpus, exact
- canonical output byte-identical, and produced by `quoin-store`'s function rather than a local one
- IO behind one trait; the analysis functions take in-memory inputs
- clippy clean at `-D warnings`

---

kreneskyp:
Ports `src/graph-analysis/` — 1,373 lines across five files — to
`rust/crates/quoin-graph-analysis`, the Stage 5 blocker for Wave 10.

## What landed

3,559 lines of Rust in eighteen modules, plus 957 lines of tests and a
one-time oracle capture.

```
src/lib.rs                     66
src/error.rs                  224   QGA-1001..1003, one thiserror enum
src/ids.rs                    304   newtypes, Ord from cmp_utf16
src/input.rs                  100   the three input contracts
src/json.rs                   268   the ONE canonical bridge
src/load.rs                   167   the ONE filesystem seam
src/render.rs                 249   canonical JSON + markdown
src/analysis/mod.rs            28
src/analysis/draft.rs         233   base()/finish(), gaps
src/analysis/fan_out.rs        68
src/analysis/churn.rs         155
src/analysis/impact.rs        312
src/analysis/paths.rs         170   shortestReversePaths + its heap order
src/model/mod.rs               44
src/model/premises.rs         225
src/model/audit.rs            309
src/model/binding.rs          249
src/model/report.rs           388
```

Largest module 388 lines; nothing over the 500-line soft ceiling, so
`tests/tc_385_module_sizes.rs` ships with both
`ALLOWED_ABOVE_HARD_CEILING` and `OVER_SOFT_CEILING` empty.

## Acceptance

* **Fan-out, churn and change impact identical to the TypeScript, exact.**
  `tests/goldens/graph-analysis.json` was captured **once**, from
  `src/graph-analysis/`, at `c55d256`: 21 analysis cases × 3 views × 2
  renderings, 30 input-contract cases, 13 loader cases. Both the canonical
  JSON and the markdown are compared **byte for byte**.
* **Canonical output produced by the store's canonicalizer.**
  `quoin_store::canonical_json_bytes` writes every byte and
  `quoin_store::canonicalize_jcs` computes every sort key.
  `tests/tc_385_one_canonical_json.rs` reads the source tree and asserts
  that those calls happen in `src/json.rs` and nowhere else, and that no
  module has grown a `to_string_pretty`, a `PrettyFormatter` or a member
  sort of its own. It carries a floor, so it fails rather than passing on
  an empty tree.
* **IO behind one trait.** `GraphInputReader` has one method, `read`. The
  retained `existsSync` probe is gone: absent-vs-unreadable is
  `ErrorKind::NotFound` on the read itself, which is the same distinction
  one syscall earlier and without the window between the two calls.
* **clippy clean at `-D warnings`**, `cargo fmt --check` clean,
  `cargo test --workspace` green.

The capture script lives under `oracle/` rather than `examples/` so it is
visibly the thing FR-101-AC-5 deletes at the Stage 5 cutover commit. No
test runs node; the TypeScript is not a runtime oracle.

## Ordering

Every `compare()` in the retained source is JavaScript `<` on strings —
UTF-16 code-unit order, which is not Rust's `str: Ord` above the BMP. Every
id newtype takes its `Ord` from `quoin_store::json::order::cmp_utf16`, so a
`BTreeMap` keyed on one *iterates* in the order the oracle sorts in rather
than being sorted afterwards. `GraphGap` and the churn event key carry
hand-written `Ord` for the same reason: `BTreeSet<GraphGap>` **is**
`uniqueGaps`, and `BTreeMap<EventKey, _>` **is** `eventOrder`.

## Declared divergence (one, and it fires)

`stableKey` (`input.ts:274`) is `JSON.stringify(canonical(v))`. That equals
RFC 8785 JCS for every document this crate sorts **except** one carrying a
member name that is an ECMAScript array index: a JavaScript object emits
integer-like names first, in ascending numeric order, whatever order they
were inserted in. This crate calls `quoin_store::canonicalize_jcs` rather
than grow a third serializer to reproduce a property-order rule.

The corpus carries a rung built to see it —
`audit/array-index-passthrough-order`, two findings alike but for
passthrough members named `2` and `10` — and `DECLARED_DIVERGENCES` fails
if that case ever stops disagreeing.

## Not a divergence, and why

`canonicalizeBindings` (`load.ts:417`) sorts on six levels; the last two
compare `JSON.stringify(symbols)` and `JSON.stringify(affirmations ?? [])`,
in JavaScript insertion order over `.passthrough()` objects. Reproducing
that would mean carrying every undeclared member of every binding, and its
byte order on disk, to break a tie that decides nothing.

It decides nothing because binding order is not observable: fan-out
accumulates into sets, churn keys events by `(obligation, who, commit,
note)`, change-impact reads only `suite` from the first binding of each
pair. `tests/tc_385_binding_order_is_not_observable.rs` asserts that
directly, permuting every corpus case's bindings three ways and requiring
all three views to come out identical. That is a stronger statement than
the tie-break, and unlike a `DECLARED_DIVERGENCES` entry it can fail — an
entry for a divergence with no observable effect could never fire.

Refusal **wording** is not compared, per quoin#403 and the `tc_470`
precedent: the verdict and the named input are contractual, the prose is
not.

## Defects found in already-merged code, fixed here

Both are reported so they can be noted on the tickets that shipped them.

1. **quoin#474 — `quoin_quire::model` re-exported half a type graph.** The
   list named `AssuranceArtifact`, `AssuranceLocator`,
   `AssuranceModulePremise`, `AssuranceObligation`, `AssuranceRelation`,
   `AssuranceSchemaPremise` and `AssuranceSymbol`, but not
   `AssuranceRelationKind`, `AssuranceRelationObservation`,
   `AssuranceResolution`, `AssuranceFreshness`, `RelationAvailability`,
   `RelationKindAvailability` or `RelationKindSource` — all reachable from
   `AssuranceExport`'s public fields. A consumer could hold an export and
   not name the enum in its `relations`, forcing it past the module the
   re-export exists to be. Fixed by completing the list.

2. **quoin#384/#385 — a closed `Severity` would have made the port
   stricter than the source.** This branch first declared
   `Finding::severity` as a closed `low|medium|high` enum, matching
   `src/auditor/audit.ts:34`. `quoin-assurance`'s captured golden corpus
   carries a finding with `severity: "error"`, which the retained
   implementation reads without complaint — TypeScript's types are not
   runtime checks, and `build_case` never reads the field. A reader
   stricter than the one it replaces refuses input that already exists, so
   `severity` travels in `other` like every other undeclared member and
   the closed vocabulary is stated exactly once, at the graph view's own
   audit contract (`src/graph-analysis/input.ts:52`), where the zod enum
   states it.

## Naming

`analysis.ts:523 selectedCorpusEdges` and `:529 AssuranceCorpusRelation`
use "corpus" for **spec-corpus graph edges**, not evidence-corpus
accounting. `analysis::impact::selected_corpus_edges` carries that warning
at its own definition.

## Not touched

No TypeScript. `src/graph-analysis/` is unchanged, `corpus/` is unchanged,
no pin weakened, no lint relaxed, no test deleted.

Closes #385


kreneskyp:
## Review — approved, merging

Gates re-run independently at `b16d390` on a clean worktree: `CLIPPY_EXIT=0`, `TEST_EXIT=0`, 123 test binaries, zero `.ts` files. 18 modules, largest 388 lines, **both** census allow-lists empty.

This ticket is what W10 (#477) was blocked on.

### The `severity` near-miss is the most valuable thing here

The port started by declaring `Finding::severity` as a closed `low|medium|high` enum, matching `src/auditor/audit.ts:34`. The workspace gate then refused a finding `build_case` has always accepted. I measured the retained data myself rather than take the report: it carries **`error` ×21, `advisory` ×4, `warning` ×3** — **not one value from the TypeScript's declared union**. Closing that vocabulary would have refused every retained finding in the assurance corpus.

A TypeScript type is not a runtime check, and `build_case` never reads the field. `severity` travelling in `other`, with the closed vocabulary stated once at the graph view's own audit contract where a zod enum actually enforces it, is the correct resolution: the retained data wins over the declared type. I have relayed this hazard to the Stage 5 auditor wave, since `audit.ts:34` is where that enum originates.

### The #474 re-export gap, fixed rather than worked around

`quoin_quire::model` re-exported only the part of the type graph a caller had needed first, so a consumer could hold an `AssuranceExport` and not name the enum inside its own `relations` — it would have had to reach past the module the re-export exists to provide. Seven types added, nine lines, with the reason in a comment. Minimal and in the right place. Noting it on #474.

### Declaring versus proving, decided correctly in both directions

**Declared (one, and it fires):** `stableKey` is `JSON.stringify(canonical(v))`, which equals RFC 8785 JCS except where a member name is an ECMAScript array index — JS emits integer-like names first and ascending regardless of insertion order. Calling `quoin_store::canonicalize_jcs` rather than growing a third serializer for that one rule is right, and the corpus rung `audit/array-index-passthrough-order` makes the must-have-fired assertion real rather than nominal.

**Not declared, deliberately:** `canonicalizeBindings`' last two tie-break levels compare `JSON.stringify` over passthrough objects in insertion order. The PR's reasoning is exactly right and worth quoting: *a divergence with no observable effect could never fire, and a declaration that cannot fire is a claim about nothing.* So instead of an inert register entry, `tc_385_binding_order_is_not_observable` permutes every case's bindings three ways and asserts all three views come out byte-identical — a property asserted where it **can** fail. Proving unreachability beats declaring it, and this is the second wave in a row to make that call correctly.

### Corpus

`tests/goldens/graph-analysis.json`, captured once from the TypeScript at `c55d256` and frozen: 21 analysis cases × 3 views × 2 renderings (canonical JSON **and** markdown, both byte-compared), 30 input-contract cases, 13 loader cases. No test spawns node (FR-101-AC-5). One filesystem seam behind one trait; the analysis functions take in-memory inputs.

No TypeScript touched, `corpus/` untouched, no pin or lint weakened.

PR #493, opened 2026-09-13
feat(rust): port the governed graph portfolio (wave 9)

### LINKED ISSUE #476: [Stage 6] quoin-measurement-graph projection: the FR-067 governed graph portfolio

kreneskyp:
Closes #476

Wave 9 of the Stage 6 measurement burn-down. Ports
`src/measurement/graph-portfolio.ts` (881 lines, FR-067) into
`quoin-measurement-graph`, the crate quoin#475 created in #491.

## What landed

15 new modules (4,333 lines of Rust including tests), largest 465 lines,
all under the 500-line soft ceiling. `ALLOWED_ABOVE_HARD_CEILING` and the
soft-ceiling allow-list both stay as W8 left them.

| retained TypeScript | here |
| --- | --- |
| `parseGraphPortfolioMappings` | `mapping.rs` |
| `buildGovernedGraphPortfolioFrom` | `build.rs`, `history.rs`, `scorer.rs` |
| `compareGraphQualityCollections` | `compare.rs` |
| `canonicalGraphPortfolioJson` | `render_json.rs` → `canonical.rs` |
| `renderGovernedGraphPortfolio` | `render.rs`, `text.rs` |
| availability / gap vocabularies | `availability.rs`, `reading.rs` |
| `compare`, `compareInstants` | `order.rs` |

## Gates

**Byte-identical on a golden corpus.** `tc_476_001` rebuilds the governed
portfolio from a committed 16-file fixture tree and compares both the
canonical JSON and the rendered Markdown byte for byte against
`tests/goldens/graph-portfolio-oracle.json` (96 KB) — captured **once**
from the retained TypeScript by `oracle/capture-graph-portfolio-oracle.mjs`,
which records its own digest, the revision it ran at, and the node version
(FR-101-AC-11). No test spawns node; `tc_476_022` asserts that as a census
over `tests/`.

**One canonical-JSON implementation.** `tc_476_020` asserts the only module
in the crate containing a canonical-JSON or pretty-print call is
`canonical.rs`, that it delegates to `quoin_store::canonical_json`, and
that it contains no hand-rolled key sort or indent (FR-100-AC-8).
`tc_476_021` does the same for the base64 alphabet, the SHA-256 hasher and
the `JsonValue` ↔ `serde_json` bridge.

Anti-vacuity floors on every case list: 15 mapping cases, 7 comparisons, 6
repositories, 4 test files, 2 `unknown` partitions.

## How this reconciles with wave 8 (#475)

- Extends `GraphAdapterError` / `GraphAdapterErrorCode` with four mapping
  codes. No second error enum.
- Uses `crate::canonical` and `crate::base64::decode` as they landed. No
  second canonical-JSON call, no second codec, no second JSON bridge.
- **Extends `tests/tc_475_module_sizes.rs`** to cover both waves rather
  than adding a `tc_476_module_sizes.rs`. One census per crate.
- W8's `tc_475_052` census broke when the four new codes widened
  `retained_spelling()`. Fixed by naming them and pointing at the file
  that does cover them, not by weakening the assertion.

## Divergences

Adds `rust/crates/quoin-measurement-graph/DIVERGENCE.md`, matching
`quoin-measurement/DIVERGENCE.md` §13: it states what the fixture covers
**and what it cannot**.

- **§2** A non-string `dimensions.measure`/`dimension`/`key` becomes
  `unknown` here, where the retained module carries the raw value into
  `markdownCell` and throws. Not in the fixture (there is no oracle for a
  throw); bounded by a corpus census that `tc_476_024` **asserts** rather
  than recites — 50,882 dimension values across 14,644 observations in 48
  collection files under `spec/evidence/measurements/`, of which 0 are
  non-strings. The **absent** case is covered: two dimensionless observations put an
  `unknown | unknown | unknown` partition inside the compared bytes, with
  a floor so it cannot quietly stop firing.
- **§3** The quoin#465 base64 leniency divergence, **widened not
  re-declared**: the actual `Buffer.from(x, "base64")` decode call site is
  `graph-portfolio.ts:786`, in this wave's module.
- **§4** Instants read by `Rfc3339DateTime`, with `Date.parse`'s `NaN`
  fall-through reproduced rather than diverged.
- **§5** The golden compares after one `@@TREE@@` path substitution.
- **§6** Divergences inherited from the plain portfolio are declared where
  they live, not restated.
- **§7** Two differences proved unreachable rather than declared: the
  `"graph partition index lost its sample"` throw (made unrepresentable by
  one merged `Pair` index) and W8's identity member ordering.

## Defect found in already-merged code

**quoin#475 landed no `DIVERGENCE.md` for `quoin-measurement-graph`** —
the only ported crate without one. Its divergences were declared in module
docs and a test table only. Fixed here (the file covers both waves) and
raised rather than absorbed.

Separately: quoin#476 and quoin#475 both cite "107 zod declarations" in
these files. `graph-portfolio.ts` declares **no zod at all**.
`tc_476_023` asserts the measured census instead: 5 exported functions, 22
private functions, 17 exported shapes, 0 zod constructors.

## Gate output

```
cargo fmt --all --check                                 FMT=0
cargo clippy --workspace --all-targets --locked -- -D warnings   CLIPPY=0
cargo test --workspace --locked                         TEST=0   1020 passed, 0 failed
```

No TypeScript was touched; cutover is W11 and deletion is W12.


kreneskyp:
## Review — approved after one cycle, merging

Gates re-run independently at `cab0e9e`: `CLIPPY_EXIT=0`, `TEST_EXIT=0`, 131 test binaries, zero `.ts` files.

### The finding, and why it mattered more than its size

As first pushed, §2 bounded a **non-fixturable** divergence — non-string `dimensions` render as `unknown` here while the TypeScript throws, so there are no bytes to compare — with a corpus census that was wrong by roughly 11×, and that **no test evaluated**. The gates were green at `d2e556a` precisely because nothing could check it. A bound that lives in prose is a bound the gate cannot hold, and this one drifted inside the commit that wrote it.

The worse half was the clause *'17 of those observations are `graph_quality`'*. There are **zero** in the retained corpus; the only ones in the repository are this wave's own fixtures. §2 exists to bound with retained data, and it was quoting the PR's own.

### The fix closes it properly

All four figures now match my independent measurement — 48 collection files, 14,644 observations, 50,882 dimension values, **0** non-strings. There is a cross-check worth recording: §2 also reports **14,339** observations stating a `dimensions` object, and I had separately measured **305** dimensionless observations while reviewing #473. 14,339 + 305 = 14,644. Two measurements taken on different days for different reasons close exactly.

`tc_476_024_no_retained_dimension_value_is_a_non_string` now **measures** what §2 asserts: a non-recursive `read_dir` over that one directory, with a comment saying why recursion is wrong here (the repository holds checkouts of itself under `.worktrees/`, which is what inflated the original count), floors an order of magnitude below the real totals, and a capped offender list. It asserts the **zero** rather than the totals, since the corpus grows — the right choice.

Mutation-checked both directions: floor raised to 100,000 fails and prints the real `48 / 14644 / 50882`; a real key treated as an offender fails the zero. That output can only come from a test genuinely reading that directory.

And §2 now states the opposite of its old claim in as many words: *'the retained corpus contains no `graph_quality` observation at all … a bound quoting those would be a self-fixture claim rather than a measurement.'* Naming the error in the register is better than quietly deleting it.

### The rest of the wave

4,333 lines of Rust, 15 modules, largest 465, under the soft ceiling.

Reconciliation with W8 is the part I would have expected to go wrong and did not: it **extended** `tc_475_module_sizes.rs` rather than adding a second census, **widened** W8's quoin#465 base64 entry rather than re-declaring it — correctly, since the real decode call site is `graph-portfolio.ts:786`, in this wave's module — and when its four new mapping codes broke W8's `tc_475_052`, it fixed that by **naming** them and pointing at the file that covers them, not by loosening the assertion.

§7 keeps two differences out of the register by proving them unreachable rather than declaring them. That is the third wave running to make that call correctly, and it is the right instinct: a declaration that cannot fire is a claim about nothing.

Also fills the `DIVERGENCE.md` that #475 never landed — noted on that ticket.

PR #561, opened 2026-09-19
FR-068: a refused change-assurance document exits 2, not 1 (#543)

### LINKED ISSUE #543: Native verify-receipt returns exit 1 instead of 2 for edited receipt
## Regression

The native Rust CLI rejects a semantically edited verification receipt, but exits 1. The documented and previously enforced contract is exit 2 when the receipt document itself is refused. This breaks tl-mltl's tamper-refusal assurance scenario: exit 1 is reserved for a well-formed receipt whose verified outcome is invalid or incomplete.

## Exact command

```sh
jq '.outcome = "invalid"' tests/fixtures/change-assurance-cli/receipt.json | quoin change-assurance verify-receipt --input - --json
```

Working directory: `/home/peter/dev/quoin`

## Expected

- Receipt is refused because a semantic field changed without recomputing its digest.
- Process exit status is 2, matching `quoin change-assurance verify-receipt --help`: “Exit status is ... 2 when the document itself is refused.”

## Actual

- Process exit status is 1.
- Full output:

```text
    Error: native runtime change_assurance.verify_receipt failed: CORE_BAD_REQUEST: invalid verification receipt: outcome disagrees with reason precedence
```

The unedited control succeeds:

```sh
quoin change-assurance verify-receipt --input tests/fixtures/change-assurance-cli/receipt.json --json
```

```json
{"candidate_revision":"candidate-1","digest":"1804806b390882babce092f6f1671438e3ed0e2e1ef6593650155276e1a24da2","outcome":"valid","reasons":[],"record_digest":"1972d0d5d3ee3fb06a2a8ca2cc97f3e239d956958368d209ebbe0a52f9b61630"}
```

## Version and platform

```text
quoin --version: 0.23.1-87-gb3d02d4
OS: Ubuntu 22.04.5 LTS under WSL2
Kernel: Linux 6.18.33.2-microsoft-standard-WSL2
Arch: x86_64
```

## Downstream impact

In agent-ix/tl-mltl, `scripts/assurance_chain.py` edits a real emitted receipt and requires `verify-receipt` to return 2. Native Quoin returns 1, so TC-022's tamper scenario fails even though the refusal message is otherwise correct.

kreneskyp:
Closes #543.

## What was wrong

Every refusal on the `quoin change-assurance` surface exited **1** instead of the documented **2** after the TypeScript-to-Rust port (#455, #457, #524).

`core_runtime::invoke` flattens a typed `CoreError` into a `String`, and `main.rs` gave every command failure the shell's generic `EXIT_COMMAND_FAILURE`. So "the document is refused" and "the receipt verified `invalid`" — the one distinction FR-068's 0/1/2 grammar exists to carry — became indistinguishable by status.

Nothing was ever accepted. A tampered receipt is refused correctly on both paths; only the status was wrong. But `agent-ix/tl-mltl` and agent-ix/quire-contract-runtime#32 both assert exit 2 on a tampered receipt, and both went red.

## The fix

The change-assurance route maps a command failure to `EXIT_DOCUMENT_REFUSED` (2). The message text is unchanged, and a receipt that verified `invalid` or `incomplete` still returns through `Outcome::Partial` at 1.

No requirement changed: this is what FR-068 already states, for every subcommand in the family — AC-1, AC-6 and AC-8 each name exit 2, and the Error Conditions section names it for a missing, unreadable, or non-JSON input.

## Why it shipped

**Its test was deleted, not never written.** `tests/change-assurance-command-surface.test.ts` carried TC-1322, TC-1324, TC-1326 and TC-1327 and was removed in the cutover with no native replacement, while the matrix kept reading ✅ against a file that no longer existed.

`quoin-cli`'s new `tc_1322_change_assurance_exit_grammar.rs` restores TC-1322 and the refusal half of TC-1324 against the real binary:

| case | status |
|---|---|
| sealed valid receipt | 0 |
| intact receipt, outcome `invalid` / `incomplete` (re-sealed, so only the verdict differs) | 1 |
| edited `.outcome` — precedence disagreement, short-circuits before hashing | 2 |
| edited `.candidate_revision` — self-consistent, only the seal is wrong | 2 |
| unparseable input; unknown `schema --name` | 2 |

Both refusal paths are covered deliberately: a fix to either one alone would have left the other reporting the verdict grammar's exit 1. The suite is red on `origin/main` and green with this change (verified by reverting the `main.rs` hunk).

TC-1326, TC-1327 and the byte-identical half of TC-1324 have no native replacement and are now recorded in the matrix as unbacked, rather than claimed against a deleted file. Restoring them is separate work.

## Gates

- `make rust-lint` — clean (fmt + clippy `-D warnings`)
- `make test` (= `rust-gate`: lint, `cargo deny`, tests) — 164 test binaries, 0 failures
- `quire validate spec/matrix.md` — 16 assert findings before and after; no new validation debt


kreneskyp:
## Review findings addressed

**High — usage errors still exited 3.** The first commit routed *dispatch* failures to 2, but the parser refuses a malformed invocation before dispatch, so `--bogus` and a flag missing its value both still hit the root shell's `EXIT_INVALID`:

```
quoin change-assurance verify-receipt --input - --bogus   # was 3, now 2
quoin change-assurance verify-receipt --input             # was 3, now 2
```

FR-068 groups usage with parse and integrity errors at 2, so the surface was answering 2 for everything the engine refused and 3 for everything the parser refused — half of AC-6.

`routed_family` now resolves the named family through Clap's own error-tolerant pass, so a parse error can be given that family's status. A hand-rolled scan for the first non-flag token would disagree with the real parser about global flags and read `--config-root change-assurance` as this surface.

`tc_1322_a_usage_error_on_the_surface_exits_two` covers all four parser refusals **and pins the negative**: `catalog list --bogus` still exits 3, so this is FR-068's grammar for one family, not a new global meaning for 2. That also retires the "TC-1322 row overclaims" finding — the usage third is now both true and tested.

**Medium — spec/matrix accuracy**, all three fixed:
- `spec/matrix.md` FR-068 coverage row said `✅ Covered` while its criteria row said `⚠️ Partial`. Now `⚠️ Partial` in both.
- FR-068's `evidence:` frontmatter still pointed at the deleted `tests/change-assurance-command-surface.test.ts` — the exact failure the log entry faults. Repointed at the new Rust test.
- `tc_457_change_assurance_boundary.rs`'s module doc said that file "is the side that still owns" the exit grammar. Corrected.

**Low** — dropped the unused `clippy::unwrap_used` allow; added `/// Provenance: quoin#543` to every test per the repo's rust-style doc. Kept the `tc_1322_` naming: `tc_1650_` in `tc_shell_contract.rs` sets the TC-id precedent in this same crate. Left `document_refused` as its own constructor.

**Not changed:** the re-derivation caveat on `reseal()` is real and accepted — a one-sided drift between sealer and verifier fails loudly, and only a coordinated drift in both would pass, which no cheaper construction avoids without hardcoding a digest that the next fixture change silently invalidates.

Gates re-run on `5775cbc`: `make test` (lint, `cargo deny`, full suite) green; `quire validate spec/matrix.md` still 16 assert findings, unchanged from `origin/main`.

